PR #49, opened 2026-05-28
[Enhancement] Accommodate calling app via a sub-path 

pixiekat:
I needed to use exact behind a reverse proxy and the static files being hard coded caused a problem with Apache serving up the proxy at /exact so this is a relatively small edit to allow for an EXACT_SUB_PATH variable which then gets consumed inside of STATIC_URL.

I also did the same for ctomop but that was more complex and I will do a seperate one for that in the appropriate repo.


LeoMo42:
Thanks for this — subpath mounting via `FORCE_SCRIPT_NAME` + WhiteNoise is the right approach and a real need. Found a few issues during review that need addressing before merge.

## Required fixes

### [P0] Dockerfile CMD is now broken — `Dockerfile:34`

The PR changes `CMD ["/app/docker/entrypoint.sh"]` → `CMD ["/docker/entrypoint.sh"]` with the comment "Fixing since it was broken at /app". The original was correct.

`COPY . /app` puts the repo under `/app/`, so the entrypoint lands at `/app/docker/entrypoint.sh`. The new path `/docker/entrypoint.sh` does not exist in the container — bare `docker run` and any orchestrator using the image CMD (Cloud Run, Kubernetes) will fail with `exec: "/docker/entrypoint.sh": no such file or directory`.

The reason this went unnoticed locally: `docker-compose.yml:35` hardcodes `entrypoint: ["/app/docker/entrypoint.sh"]`, overriding the Dockerfile CMD. So `docker compose up` works fine; the built image does not.

**Fix:** revert to `CMD ["/app/docker/entrypoint.sh"]`.

---

### [P1] `SUB_PATH` not validated — `exact/settings.py:167`

`os.environ.get('SUB_PATH', '').rstrip('/')` only strips trailing slashes. Two input cases cause silent breakage or worse:

**Missing leading slash** (`SUB_PATH=exact`): `STATIC_URL` becomes `'exact/static/'` (a relative URL). Django requires `STATIC_URL` to start with `/`. All static assets 404, `reverse()` produces malformed URLs (`exactadmin/`), and no error is raised at startup.

**CRLF in value** (`SUB_PATH=/exact\r\nX-Injected: evil`): passes through `rstrip('/')` unchanged. `FORCE_SCRIPT_NAME` carries the CRLF. Django uses the script prefix in login redirect `Location` headers — this is a header injection class of issue.

**Suggested fix:**

```python
SUB_PATH = os.environ.get('SUB_PATH', '').strip().rstrip('/')
if SUB_PATH:
    if not SUB_PATH.startswith('/') or any(c in SUB_PATH for c in ('\r', '\n', '..')):
        from django.core.exceptions import ImproperlyConfigured
        raise ImproperlyConfigured(
            f"SUB_PATH must be an absolute path without CRLF or path traversal (got {SUB_PATH!r})"
        )
```

---

## Worth addressing / investigating

**`exact/wsgi.py` rationale** — The explicit `django.setup()` before `get_wsgi_application()` is non-standard. The comment says "get_wsgi_application() calls django.setup(set_prefix=False)" — Django actually calls it with `set_prefix=True`. More importantly, WhiteNoise's `__init__` runs when Django constructs the middleware chain (first request), not at `get_wsgi_application()` call time, so the ordering may not have the effect described. The double `setup()` is idempotent and harmless, but please verify with a real gunicorn boot that `/exact/static/...` is actually served correctly when `SUB_PATH=/exact` before shipping.

**Env var name mismatch** — The PR description mentions `EXACT_SUB_PATH` but the implementation reads `SUB_PATH`. Worth clarifying which name is canonical (the namespaced `EXACT_` prefix is a reasonable convention for this project).

---

*Reviewed by: Claude + Codex (independent adversarial passes — all three sources agree on the P0 and P1 findings above).*

Issue #108, opened 2026-05-31
Bind CTOMOP person_id resolver to authenticated caller (IDOR follow-up to #102)


LeoMo42:
Follow-up to #102 (PR #107) and part of #101's identity-flow scope.

## Background

PR #107 added a server-side `?person_id=` resolver that fetches the patient record from CTOMOP. The current implementation:

- Calls CTOMOP with a static service token (`CTOMOP_SERVICE_TOKEN`).
- Does **not** verify that the authenticated DRF user is authorized to access the requested `person_id`.

Automated security review correctly flagged this as IDOR.

## Why we shipped it this way

EXACT is stateless for patient data (per project memory `feedback_exact_no_own_db.md`) — there's no model linking users to patients in EXACT's own DB, so there's no in-tree relationship to verify against. Adding fake EXACT-side authz (e.g. a hard-coded `is_staff` check) would be worse than nothing.

The boundary is documented in `trials/services/patient_info/resolve.py`'s module docstring: production deployments MUST either enforce per-user authz at the CTOMOP layer, or swap the static service token for caller-identity forwarding (e.g. exchanging the user's token rather than using a static service token).

## Scope of this issue

Pick one (or both) of:

1. **Identity forwarding** — Replace `CTOMOP_SERVICE_TOKEN` with caller-identity passthrough so CTOMOP enforces row-level authz against `request.user`'s real identity. Likely shape: an OAuth token exchange flow or signed JWT propagation. Coordinate with #101's federation identity model.

2. **EXACT-side authz** — If EXACT adopts a per-user patient-relationship model (e.g., clinician → person mapping), wire that into `_resolve_from_ctomop` before the network call.

## Acceptance

- A user authenticated as person A cannot retrieve trials for person B by guessing/scraping `person_id`s.
- The `CTOMOP_SERVICE_TOKEN` is either removed or restricted to a low-privilege bootstrap role.
- Documented in `resolve.py` once the design lands.

## Risk if deferred

Production deployments that haven't enforced CTOMOP-side row-level authz will leak patient PHI across users. Status quo is documented but load-bearing on operator vigilance.

## Effort
M (depends on which path #101 picks).

LeoMo42:
[...]

PR #116, opened 2026-05-31
fix(frontend): TrialCard meta line stray leading separator

LeoMo42:
## Bug

`/trials/` list (without patient context) returns `matchScore=null` but `goodnessScore=25`. The card's meta line used naive `&&`-conditional concatenation:

\`\`\`tsx
{trial.matchScore != null ? \`Match score: \${trial.matchScore}\` : null}
{trial.goodnessScore != null ? \` · Goodness: \${trial.goodnessScore}\` : null}
\`\`\`

When matchScore is null and goodnessScore is set, the second chunk's leading \` · \` is the first thing rendered → cards show ` · Goodness: 25` with a stray dot.

## Fix

Collect present pieces into an array and \`join(' · ')\` between them. Same pattern applied to the studyId/status/phase line above for symmetry. The whole meta \`<div>\` is hidden when no pieces are present.

## How surfaced

During a local demo run of the federation harness against a local EXACT instance. Patient context was silently missing because the local EXACT's `CTOMOP_BASE` server-to-server resolver wasn't wired with credentials → matcher returned matchScore=null → leading-dot bug fired on every card.

The underlying server-to-server CTOMOP-auth gap is a separate concern (effectively blocked on #108's identity flow). This PR only fixes the rendering.

## Test plan

- [ ] CI: `typecheck` + `build` pass.
- [ ] Local: cards with patient context show `Match score: 100 · Goodness: 25`.
- [ ] Local: cards without patient context show just `Goodness: 25` (no leading dot).
- [ ] Local: a trial with no scores at all hides the meta line entirely.

### DIFF
--- a/frontend/src/federation/TrialCard.tsx
+++ b/frontend/src/federation/TrialCard.tsx
@@ -55,14 +55,28 @@ export function TrialCard({ trial, onSelect, isSelected }: Props) {
         <span style={{ ...parseStyle(m.badge) }}>{m.label}</span>
       </div>
       <div style={meta}>
-        {trial.studyId} · {trial.recruitingStatus}
-        {trial.phase && trial.phase.length ? ` · ${trial.phase.join(" / ")}` : null}
+        {[
+          trial.studyId,
+          trial.recruitingStatus,
+          trial.phase && trial.phase.length ? trial.phase.join(" / ") : null,
+        ]
+          .filter(Boolean)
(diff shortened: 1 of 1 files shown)


LeoMo42:
Folded into PR #117 (feat/harness-ctomop-inline-fetch) — the card meta-line fix and the harness CTOMOP-inline-fetch travel together for the live demo.

Issue #102, opened 2026-05-31
feat(patient-info): server-side CTOMOP→EXACT patient resolver (person_id fetch + field map)


LeoMo42:
Part of #101 (Track A).

## What
Add a CTOMOP-sourced path to `trials/services/patient_info/resolve.py` so a caller can pass `?person_id=` instead of a full inline `patientInfo`, and EXACT fetches the record from CTOMOP and maps it onto its `PatientInfo` class. Mirrors SoC's `core/pipeline/patient_adapter.py`, server-side.

## Tasks
- [ ] Settings: `CTOMOP_BASE` + service token; httpx client (cf. hk-labs `ctomop_client.py`).
- [ ] `resolve_patient_info`: when `person_id` is present, `GET {CTOMOP_BASE}/api/patient-info/{id}/` and map; otherwise keep the existing inline path **unchanged**.
- [ ] Field map: CTOMOP flat record → EXACT `patientInfo` — disease, patientAge, gender, ECOG/Karnofsky, ~16 labs, geneticMutations, therapy lines + outcomes, stemCellTransplantHistory, ER/PR/HER2/menopausal, Binet, FLIPI inputs.
- [ ] Reuse `exactomop/OMOP_TO_PATIENTINFO_MIGRATION.md` mapping where it overlaps; reconcile differences with ctomop's projection.
- [ ] Unit tests against ctomop seeded patients (person_ids 9001–9007).

## Notes / constraints
- **Do not add Django model fields** for patient data — declare on the Python class (`patient_info.py:237`) + `configs.py`/`normalize.py`. No migrations.
- Keep the inline-JSON contract intact (CancerBot depends on it).
- Confirm patient source is **ctomop** (the picker source), not `exactomop`.
- Output keys are camelCase end-to-end.

PR #120, opened 2026-06-01
fix(harness): axios v1 XHR adapter drops GET body — switch to fetch adapter

LeoMo42:
## Bug

EXACT's \`/trials/\` reads patient context from the request body even on GET (matches the existing CB inline contract used in #102's \`resolve_patient_info\`). When the harness (per #117) sends \`patient_info\` inline via:

\`\`\`ts
apiClient.get('/trials/', { params, data: body })
\`\`\`

**axios v1's default XHR adapter silently drops the body on GET requests** — Django receives the URL with no \`patient_info\`, \`resolve_patient_info\` returns \`None\`, the matcher runs with no patient context, and the response is the disease-agnostic union.

Result: a CLL patient (#9006) shows MM trials in the result list.

## Root cause

Verified end-to-end against EXACT's runserver: every harness request hit Django as \`GET /trials/?person_id=… HTTP/1.1\` with \`Content-Length: 0\`. The body never left axios's XHR adapter.

## Fix

Set \`adapter: "fetch"\` on the EXACT axios instance (\`exactAuth.ts:makeExactClient\`). The fetch adapter honours GET-with-body and the payload reaches Django intact.

\`CtomopClient\` (\`src/dev/ctomopClient.ts\`) is deliberately left on the default XHR adapter because it only uses POST + GET-without-body. XHR is fine for those, and switching it would require re-validating session-cookie behaviour through the dev proxy.

## How surfaced

User caught the discrepancy during the local demo run that's also producing #117 (harness inline-fetch), #118 (federation axios shim), and #119 (recruitment dropdown).

## Test plan

- [ ] CI: \`typecheck\` + \`build\` pass.
- [ ] Local: CLL patient #9006 in the harness returns only CLL trials, not MM.
- [ ] Local: EXACT runserver log shows \`GET /trials/ HTTP/1.1\` with patient_info body bytes (no longer \`Content-Length: 0\`).

### DIFF
--- a/frontend/src/dev/exactAuth.ts
+++ b/frontend/src/dev/exactAuth.ts
@@ -77,5 +77,15 @@ export function makeExactClient(
     baseURL,
     timeout: DEFAULT_TIMEOUT_MS,
     headers: { Authorization: `Token ${token}` },
+    // EXACT's `/trials/` reads patient context from the request body
+    // even on GET (matches the existing CB inline contract). axios v1's
+    // default XHR adapter silently drops the body on GET — the request
+    // reaches Django with no `patient_info`, the matcher runs with no
+    // patient context, and the response is the disease-agnostic union.
+    // The fetch adapter honours GET-with-body, so EXACT actually
+    // receives the payload. (Verified end-to-end against EXACT's
+    // runserver: the XHR path logged 0-byte bodies, the fetch path
+    // logs full payloads.)
+    adapter: "fetch",
   });
 }


LeoMo42:
Doesn't actually fix it — the Fetch API spec also forbids GET-with-body, so switching axios to the fetch adapter just moves the error to construction time. Replaced by a proper POST endpoint on the backend (#121) + frontend switch (PR to follow).

Issue #128, opened 2026-06-01
Track C: GCP/Cloud Run deploy via Workload Identity Federation (replace Docker-Hub-only CD)


LeoMo42:
Part of #125.

## What

EXACT's only CD today is `.github/workflows/docker.yml` → Docker Hub (`secrets.DOCKER_IMAGE/USERNAME/PASSWORD`). There is no GCP deploy. Add a keyless Cloud Run deploy mirroring SoC `.github/workflows/deploy-staging.yml`.

## Tasks

- [ ] GitHub Actions workflow on push to `main`: keyless WIF auth (`google-github-actions/auth@v2`), no long-lived JSON key.
- [ ] Build + push image to Artifact Registry (`${GCP_REGION}-docker.pkg.dev/${GCP_PROJECT_ID}/ht-phr/exact`).
- [ ] Cloud Run blue-green: `update --no-traffic` → run migrations via a Cloud Run **job** → `update-traffic --to-latest`.
- [ ] Gate the deploy behind the existing CI (django.yml etc.) so a red build never deploys.
- [ ] **Container port 8000 → 8080**: EXACT's Dockerfile/entrypoint bind 8000; Cloud Run house default is 8080. Bind `0.0.0.0:${PORT}` with 8080 default.
- [ ] Config via GitHub Actions **variables** (not secrets): `GCP_PROJECT_ID`, `GCP_WORKLOAD_IDENTITY_PROVIDER`, `GCP_SERVICE_ACCOUNT`, `GCP_REGION`. App secrets via Secret Manager → env.

## Constraints

- Keep credentials out of the workflow file; WIF + Secret Manager only (no committed keys).
- Coordinate the Artifact Registry repo / SA with Track C-Terraform (#TBD) and `ht-phr/infra`.

## Effort
M

LeoMo42:
Done. The keyless-WIF Cloud Run deploy workflow landed via **#139** (`.github/workflows/deploy-staging.yml`: WIF auth → Artifact Registry → blue-green `update --no-traffic` → migrate Cloud Run job → `update-traffic --to-latest`), and the container port 8000→8080 fix landed via **#137**. Neither PR carried a `closes` keyword, so this stayed open. Repo-side deploy mechanism is complete; remaining work is operational (set the 5 `GCP_*` repo vars + `terraform apply`).

PR #137, opened 2026-06-02
fix(docker): bind $PORT (default 8080) for Cloud Run

LeoMo42:
## Summary
First of the #128 breakdown. Makes the container honor Cloud Run's injected `$PORT` (house default 8080) instead of a hardcoded 8000.

- `Dockerfile`: `PORT=8000` → `8080`, `EXPOSE 8000` → `8080`
- `docker/entrypoint.sh`: gunicorn `--bind 0.0.0.0:8000` → `--bind "0.0.0.0:${PORT:-8080}"`
- `docker-compose.yml`: pin container `PORT: "${PORT:-8000}"` so local dev stays on 8000 (the published mapping) while Cloud Run injects 8080

## Why the compose change
Baking `PORT=8080` into the Dockerfile would otherwise break local `docker compose up`: compose publishes `${PORT:-8000}:${PORT:-8000}`, but the container would inherit `PORT=8080` from the Dockerfile ENV and gunicorn would bind 8080 with nothing on the mapped 8000. Pinning the container's `PORT` keeps the published port and the bind port in sync locally.

## Test plan
- [ ] `docker compose up` serves on http://localhost:8000
- [ ] Image run with `-e PORT=8080` binds 8080 (Cloud Run path)

## Review note
Claude fresh-context self-review passed (caught + fixed the compose regression above). `codex review` could not run: codex CLI v0.130.0 rejects every model for ChatGPT accounts (`"... model is not supported when using Codex with a ChatGPT account"`). Codex half of the house two-part review is deferred until the CLI auth/model gating is resolved.

Part of #128.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM python:3.12-slim
 
-ENV PYTHONUNBUFFERED=1 PYTHONHASHSEED=random PYTHONDONTWRITEBYTECODE=1 PORT=8000
+ENV PYTHONUNBUFFERED=1 PYTHONHASHSEED=random PYTHONDONTWRITEBYTECODE=1 PORT=8080
 
 RUN apt-get update \
     && apt-get -y --no-install-recommends install \
@@ -28,6 +28,6 @@ COPY . /app
 
 RUN python manage.py collectstatic --no-input
 
-EXPOSE 8000
+EXPOSE 8080
 
 CMD ["/app/docker/entrypoint.sh"]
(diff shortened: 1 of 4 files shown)


LeoMo42:
**Codex review (now restored, v0.136.0):** ran clean after one fix.

- [P2] Codex flagged that the published-image quickstart in `docker/Image.md` mapped `8000:8000` against an image that now defaults to `PORT=8080`, so `localhost:8000` would no longer reach gunicorn. Fixed in dffae17 by pinning `PORT: "8000"` in the documented compose. Re-review clean.

Two-part self-review now complete: Claude fresh-context + codex both pass.

Issue #157, opened 2026-06-12
[MEDIUM] rediss:// Celery broker/result SSL disables certificate verification (CERT_NONE)


LeoMo42:
**Severity:** MEDIUM · **Source:** Codex

**Where:** `exact/settings.py:273-275`

**Problem:** `rediss://` Celery broker/result SSL uses `ssl.CERT_NONE`. If Celery ever carries PHI or patient-derived task results, this permits MITM on Redis traffic.

**Fix:** Require verified TLS (CA validation) for non-local environments.

---
_Filed from the 2026-06-12 platform security/arch audit (8-agent Claude review + Codex second opinion). Full report: `platform-review-2026-06-12.md`._

LeoMo42:
Resolved by PR #164 (merged to main): rediss Celery broker/result use ssl.CERT_REQUIRED outside local/DEBUG (CERT_NONE only for self-signed dev), with REDIS_SSL_CA_CERTS for a private CA.

PR #164, opened 2026-06-15
fix(celery): verify rediss TLS certificates outside local/DEBUG (#157)

LeoMo42:
## Summary
- The `rediss://` Celery broker and result backend pinned `ssl.CERT_NONE`, disabling certificate verification and permitting MITM on Redis traffic that can carry patient-derived task results.
- Use `CERT_REQUIRED` in deployed environments (`CERT_NONE` only for local/DEBUG self-signed dev), with `REDIS_SSL_CA_CERTS` to point at a private CA bundle.
- Source-level guard tests added.

## Review
- Codex review: no regressions; preserves local/DEBUG behavior.

## Test plan
- [ ] `manage.py check` passes
- [ ] With `rediss://` + non-local ENVIRONMENT, broker/result SSL uses `CERT_REQUIRED`

### DIFF
--- a/exact/settings.py
+++ b/exact/settings.py
@@ -271,8 +271,19 @@ redis_url = os.environ.get('REDIS_URL', 'redis://localhost:6379')
 CELERY_BROKER_URL = redis_url
 CELERY_RESULT_BACKEND = redis_url
 if CELERY_BROKER_URL.startswith('rediss://'):
-    CELERY_BROKER_USE_SSL = {'ssl_cert_reqs': ssl.CERT_NONE}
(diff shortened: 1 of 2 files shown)


LeoMo42:
⚠️ **Deploy note (from fresh-context review):** this is a hard cutover, not behind a new flag. Any deployed `rediss://` env that currently connects with a self-signed/private cert via CERT_NONE will now **fail the TLS handshake** unless it either (a) presents a publicly-trusted cert or (b) sets `REDIS_SSL_CA_CERTS` to its CA bundle. Confirm each `rediss://` deploy before rollout. (Public-CA / managed Redis works out of the box — redis-py uses the system trust store under CERT_REQUIRED.)

Issue #153, opened 2026-06-12
[HIGH] Persistent /api-token-auth/ DRF tokens; TokenAuthentication enabled globally


LeoMo42:
**Severity:** HIGH · **Source:** Codex

**Where:** `exact/urls.py:10,12,18`; `exact/settings.py:232`

**Problem:** `/api-token-auth/` issues persistent DRF tokens via username/password, and `TokenAuthentication` remains enabled globally as "transitional". For a HIPAA PHI service this creates long-lived bearer tokens with no expiry, scope, audience, tenant binding, or rotation policy.

**Fix:** Disable DRF token auth and the token endpoint in prod; use Firebase/OIDC or short-lived, scoped service credentials with audit/rotation.

---
_Filed from the 2026-06-12 platform security/arch audit (8-agent Claude review + Codex second opinion). Full report: `platform-review-2026-06-12.md`._

LeoMo42:
Resolved by PR #165 (merged to main): DRF TokenAuthentication and /api-token-auth/ are gated behind ENABLE_DRF_TOKEN_AUTH, which fails closed (off) outside local/DEBUG.

PR #165, opened 2026-06-15
fix(auth): gate DRF token auth + token endpoint to non-prod (#153)

LeoMo42:
## Summary
- Persistent DRF tokens (username/password → never-expiring, unscoped bearer, no rotation) are unacceptable for a PHI service in production.
- Add `ENABLE_DRF_TOKEN_AUTH` (defaults **on** for local/DEBUG, **off** in prod): `TokenAuthentication` is only added to `DEFAULT_AUTHENTICATION_CLASSES` and `/api-token-auth/` is only routed when the flag is on. Deployed callers use Firebase / the service token.
- Source-level + behavioral guard tests added.

## Review
- Codex review: no regressions in modified paths.

## Test plan
- [ ] In test/local env: token auth works, `/api-token-auth/` resolves (existing token-auth tests pass)
- [ ] With `ENVIRONMENT=prod` (no override): `TokenAuthentication` absent, `/api-token-auth/` unrouted

### DIFF
--- a/exact/settings.py
+++ b/exact/settings.py
@@ -214,10 +214,10 @@ DEFAULT_AUTO_FIELD = 'django.db.models.BigAutoField'
 # House API auth standard (OIDC shared-Identity model)
 # PartnerAuthentication verifies a Firebase ID token (the only
 # browser-appropriate credential); ServiceTokenAuthentication compares a shared
-# bearer against SERVICE_AUTH_TOKEN for server-to-server calls. TokenAuth +
-# Session remain as a transitional credential for the local dev harness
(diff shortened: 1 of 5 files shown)


LeoMo42:
⚠️ **Deploy note (from fresh-context review):** the federation host contract (`frontend/src/federation/types.ts`) currently documents `Authorization: Token <…>`. With DRF token auth disabled in prod, any prod federation host still authenticating EXACT with a DRF token will get **401**. Confirm all prod hosts use Firebase (PartnerAuthentication) or the service token (`Bearer`) before merge, and update the stale contract doc. The service-token path uses a different scheme (`Bearer`), so there's no automatic fallback for a Token-based host.

PR #180, opened 2026-06-19
perf: cut /match/ latency — fix all_options() + matcher N+1s, add shared Redis cache

### LINKED ISSUE #3: Ethnicity not handled correctly
## Description
Ethnicity is not handled correctly in EXACT for this trial.
These cases were flagged as missing from CB.

### Affected OMOP users :
- Hannah Anderson
- Evelyn Lopez
- Addison Harris
- Amelia Wilson

LeoMo42:
Perf fixes for the slow `POST /trials/{id}/match/` (~6s warm on staging). Three independent changes.

## 1. `Therapy.full_title()` N+1 — the main `all_options()` win

`ValueOptions.all_options()` cold build was **~488 queries / ~1.3s** (local pg). ~288 were a single N+1: `full_title()` called `self.components.order_by('id').all()`, and `.order_by()` builds a fresh queryset that **ignores** the `prefetch_related('components')` cache the therapy loaders set up — one component query per therapy, ~25 loader calls over.

**Fix:** sort components in Python so the prefetch cache is reused. **488 → 204 queries, ~1346ms → ~271ms.** Independent of cache backend, so it helps staging today.

## 2. Matcher components/categories N+1 — per-request hot path

`UserToTrialAttrMatcher.therapy_related_things_match_status()` loaded therapies without prefetch, then iterated `therapy.components.order_by('id')` and `component.categories.all()` — query count scaled with the patient's therapy count (~16 queries for 8 therapies, ~254 for the full set).

**Fix:** prefetch `components__categories` + Python sort (same pattern as #1). **Constant 3 queries** regardless of therapy count. Behaviorally identical (lists used only for dedup; result emitted as sorted(set(...))).

## 3. Shared Redis cache (latent until Redis exists)

`all_options()` is cached 1h, but `CACHES` used per-process `LocMemCache`; under `gunicorn --workers 4` on Cloud Run the cache was never shared. `build_caches()` points the default cache at Redis in deployed envs with `REDIS_URL` set, else LocMem (local/CI/tests, and deployed-without-Redis as a safe degrade). Includes a TLS-bypass fix: redis-py's `from_url()` lets URL querystring override `OPTIONS`, so `rediss://…?ssl_cert_reqs=none` would silently disable cert verification — we strip TLS-policy query params so our `CERT_REQUIRED` is authoritative.

### ⚠️ Staging has no Redis today

`gcloud run services describe exact-staging` shows **no `REDIS_URL`** (and no Memorystore). Fix #3 is **inert on staging until Redis is provisioned** — right architecture, ready for when it lands (also unblocks Celery). The latency drop now comes from #1 and #2, which are backend-agnostic.

To activate #3 later: provision Memorystore/Redis and set `REDIS_URL` on the Cloud Run service.

## Tests

- `full_title()` 0 queries when prefetched; cold `all_options()` build under a 300-query budget.
- Matcher query count is constant across therapy counts (non-scaling), stays ≤ 5.
- Cache resolution matrix, `rediss://` TLS policy, querystring-downgrade defense (end-to-end `ConnectionPool` check), source guard vs hardcoded LocMem, `all_options()` caching guard.
- Full suite green (418 passed, 5 skipped in touched areas).

## Review

Two-part self-review (fresh-context Claude + `codex review`) on each of the three commits: all findings (2 P2s on the cache commit; nits on the N+1 commits) addressed or consciously deferred with inline notes; both reviewers clean on the final tree.

### DIFF
--- a/exact/cache_config.py
+++ b/exact/cache_config.py
@@ -1,0 +1,86 @@
+"""Django cache backend resolution.
+
+Factored out of settings.py so the choice between a shared Redis cache and a
+per-process in-memory cache is unit-testable without reloading the frozen
+settings module.
+
+Why this matters (perf, #match-endpoint): `ValueOptions.all_options()` costs
+~490 sequential DB queries to build and is cached for 1h. Under the deploy
+config (`gunicorn --workers 4` on Cloud Run, multiple instances) a per-process
+`LocMemCache` is never shared, so every cold worker re-pays the full build —
+~6s per request over Cloud SQL network latency. Pointing the default cache at
+the already-provisioned Redis (the same instance Celery uses) makes the blob
+shared across all workers and instances: built once, then ~2ms per hit.
+"""
+import ssl
+from urllib.parse import urlsplit, urlunsplit, parse_qsl, urlencode
+
+LOCMEM_BACKEND = "django.core.cache.backends.locmem.LocMemCache"
+REDIS_BACKEND = "django.core.cache.backends.redis.RedisCache"
+
+# redis-py's ConnectionPool.from_url() lets querystring args OVERRIDE kwargs
+# passed via OPTIONS ("querystring arguments always win"). So a deployed
+# REDIS_URL of `rediss://host?ssl_cert_reqs=none` would silently defeat the
+# CERT_REQUIRED we set in OPTIONS, disabling TLS verification on a cache that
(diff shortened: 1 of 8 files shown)


LeoMo42:
Fix #3 (shared Redis cache) is inert on staging until Redis is provisioned — tracked in #181. Fixes #1 and #2 (the N+1s) speed up staging regardless.

Issue #256, opened 2026-07-28
chore(omop): vocab-mirror sync follow-ups — checksum verification + COPY load


LeoMo42:
Follow-ups deferred from #250 (surfaced in review), tracked so they aren't lost:

**1. Checksum verification.** The sync records the manifest `checksums` (per-table SHA-256, promop#334) on `MirrorRelease` but does **not** yet verify them — enforced completeness is the `__done` sentinel count + manifest `row_counts` (which reject truncated/partial loads). Byte-level checksum verification would additionally catch *silent corruption* (same count, altered rows). Blocked on **pinning promop's exact per-table row canonicalization** (over raw NDJSON bytes? over rows? sentinel included?) — a wrong guess would fail every sync. Confirm with promop, then verify + fail closed on mismatch.

**2. COPY for the large tables.** `concept_relationship` (~18M rows) loads via batched `bulk_create` into a table with live indexes. Move to `psycopg.copy` (optionally drop/rebuild the release's indexes around the load) for a large speedup on the periodic sync.

Both are in `vocab_mirror/sync.py` (see the module docstring notes). Neither blocks the mirror going live; the completeness gate + populated-generation activation gate keep loads safe today.

LeoMo42:
Closed by #295 (COPY-based snapshot ingestion), merged to `2omop`. Per that PR the actionable COPY half is done; the byte-level checksum-verification half stays blocked cross-repo (promop `checksums` are {count, min/max ctid} — physical ctids aren't consumer-verifiable), tracked separately. Merged to a non-default branch so it didn't auto-close. Reaches prod via #223.

Issue #250, opened 2026-07-28
feat(omop): vocab sync command (promop #334) + ops


LeoMo42:
Part of epic #246. See ADR 0002 (#245) mechanisms #6/#7.
- Implement promop #334: conditional-poll `/api/v1/vocab-releases/latest/` with the **last successfully-processed ETag** (do NOT advance the ETag on mere observation — a failed R must not become a 304); download NDJSON snapshots into **staging/inactive generation**; verify **`__done` sentinel + row_counts + checksums**; COPY-load; mark READY; hand to T4b activation.
- **Auth:** vocab-scoped OAuth client — do NOT reuse `PromopClient` unchanged (it's patient-endpoint-specific, scope `patient/*.read`); needs an explicit vocabulary scope/client.
- **Ops:** run as a **Cloud Run Job** (never in a gunicorn request); **Postgres advisory lock** = singleton writer (Redis-free); **Cloud Scheduler** cadence (6h prod / 1h staging); temp disk for ~18M `concept_relationship` rows; alerting; readiness before initial load. (Current deploy only creates the old cache table — deploy-staging.yml.)

LeoMo42:
**Schema note from #248 review (idempotency):** `MirrorConceptRelationship` / `MirrorConceptAncestor` deliberately carry **no uniqueness constraint** (~18M rows; snapshot is authoritative). So the DB will **not** catch a re-run/retry of the loader for the same `release_id` — duplicate edges would inflate descendant/component sets. This loader (#250) must therefore make same-release load **idempotent by construction**: delete-by-`release_id`-then-insert, or load into a fresh staging generation and only activate after verification — never a bare append into an existing release.

LeoMo42:
Implemented by #257, merged to the `2omop` epic branch. It merged to a non-default branch, so it did not auto-close — closing manually. Reaches prod via the 2omop→dev/main cutover (#223).

Issue #249, opened 2026-07-28
feat(omop): release contract + atomic activation + release-match gate


LeoMo42:
Part of epic #246. See ADR 0002 (#245) §Principle / mechanisms #1-#4.
- `release_manifest`(release_id, upstream ETag, checksums, row_counts, state, loaded_at, health) + single-row **`active_release`** pointer.
- **Atomic activation** = one short `default`-DB txn: verify all artifacts READY + same release_id + checksums, then flip the pointer. That DML commit IS the swap (no table rename / search_path).
- **Release-match gate:** activate R only when mirror + `TherapyOmopMapping` + `ComponentCategoryOmopLookup` + trial `omop_*` projection all agree on R (v1 = light gate on a single metadata value); mismatch → don't activate, keep last consistent bundle, alert.
- **Retention window** for the previous generation (replaces REPEATABLE-READ; keeps in-flight runs valid).

LeoMo42:
Implemented by #255, merged to the `2omop` epic branch. It merged to a non-default branch, so it did not auto-close — closing manually. Reaches prod via the 2omop→dev/main cutover (#223).

Issue #246, opened 2026-07-28
epic(omop): adopt promop vocab mirror + client-side consistency (supersedes #234 API+cache)


LeoMo42:
Umbrella for moving EXACT from **API+cache** (concept-graph query + result cache, #234) to a **release-pinned local vocabulary mirror** (promop #334), with the consumer-side consistency design in ADR 0002 (#245).

**Why:** client-side whole-vocab consistency (scoped to OMOP-derived graph expansion — types/planned stay CB-code space). See [ADR 0002](https://github.com/healthkey-ai/exact/blob/docs/adr-vocab-mirror-consistency/docs/adr/0002-vocab-mirror-consumer-consistency.md).

**Depends on:** promop revising ADR 0001 API+cache→mirror — **healthkey-ai/promop#337**.

**Sub-tickets (sequence):**
- [ ] #247 — ADR/bookkeeping fixups (T2)
- [ ] #248 — Mirror schema + placement + semantic indexes (T4)
- [ ] #249 — Release contract + atomic activation + release-match gate (T4b)
- [ ] #250 — Sync command (#334) + ops (T5)
- [ ] #251 — Local traversal + retire #234 API+cache surface (T6)
- [ ] #252 — Wire into matcher/backfill (Phase-T) + fail-closed + observability (T7)
- [ ] #253 — Re-scope #232 / mark #234 superseded (T8)

**v1 vs deferred** (ADR §'What lands now'): v1 = generation-tag + pointer-flip + per-request pin + retention window + advisory-lock sync + fail-closed + light release-match gate. Deferred = list-partitioning, REPEATABLE-READ/dual-DB, full versioned CB trial-projection contract (Conway risk).

**Guardrails:** don't obscure the #221 supportive-therapy mismatch; preserve None(unknown) vs [](known-empty) for patient components.

PR #257, opened 2026-07-28
feat(omop): vocab mirror sync loader (promop #334) (#250)

LeoMo42:
Sync slice of the vocab-mirror epic **#246** (ADR 0002). Closes **#250**. **Stacked on #255 (#249)** — base `feat/249-release-contract-activation`; review after #254/#255.

## What
`sync_vocab_mirror` management command (a scheduled **Cloud Run Job**) implementing promop's snapshot protocol (promop#334):
- **advisory-lock singleton** writer (the #249 activation control plane assumes it);
- **conditional ETag poll** of `vocab-releases/latest` (the ACTIVE release's ETag);
- stream each table's NDJSON snapshot into a **fresh staging generation** (delete-by-release-then-insert — idempotent; mirror has no uniqueness);
- **verify** completeness (`__done` sentinel count + manifest `row_counts`), **fail closed** (→ FAILED, rows dropped) on truncation/mismatch;
- mark READY and **activate via #249**.
- dedicated **vocab-scoped OAuth** client (`PROMOP_VOCAB_OAUTH_*`), distinct from the patient client's `patient/*.read`.

## Review & tests
Two-part self-review (fresh-context agent + codex) caught real issues, all fixed:
- **P1 (agent):** a READY-but-stranded activation would 304-skip forever (permanent 503). Fixed — "last processed" ETag = ACTIVE only; a READY generation self-recovers by re-activating (no re-download).
- **P1 (both):** `--tables` subset + activate could put an incomplete generation live (fail-open). Fixed — the populated-generation gate now requires rows in **all four** tables; subset loads stage without activating.
- **P2:** streamed responses closed (`with resp` + generator close + mid-stream error → typed); sentinel must carry an integer count.
- **P3:** malformed manifest → typed `VocabSyncError`.

Tests: **31** vocab_mirror (happy path, 304, already-active, **READY-recovery**, truncation/count/bad-sentinel/malformed-manifest fail-closed, **subset-not-activated**, restage dedup, advisory-lock); full suite **1073 passed / 5 skipped**; `makemigrations --check` clean.

## Deferred (tracked in #256)
Checksum verification (needs promop's row canonicalization) and COPY-based loading for the ~18M-row table. Neither blocks going live — the completeness + populated-generation gates keep loads safe.

### DIFF
--- a/exact/settings.py
+++ b/exact/settings.py
@@ -397,6 +397,20 @@ PROMOP_OAUTH_TOKEN_URL = os.environ.get('PROMOP_OAUTH_TOKEN_URL', '')
 # unconfigured (callers get ConceptGraphUnavailable) so nothing silently no-ops.
 PROMOP_API_BASE = os.environ.get('PROMOP_API_BASE', '')
 
+# promop vocabulary MIRROR sync (#250; ADR 0002 / promop#334). EXACT keeps a
+# release-pinned local mirror of the OMOP vocab tables and syncs it from promop's
+# vocab-releases snapshot API. This needs its OWN OAuth client + scope: the
+# snapshot endpoints are vocabulary-scoped, distinct from the patient client's
+# PROMOP_OAUTH_* (scope patient/*.read). Base falls back to PROMOP_API_BASE /
+# PROMOP_BASE (same promop host); token_url defaults to {base}/o/token/.
+PROMOP_VOCAB_BASE = os.environ.get('PROMOP_VOCAB_BASE', '')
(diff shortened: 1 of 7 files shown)


LeoMo42:
**Holistic whole-stack review (#248→#251) fix:** force-pushed a fix for a composition bug this slice introduced — a subset `--tables` load marked the release READY, wedging a later full sync (couldn't activate the incomplete generation via the all-tables gate, nor restage a READY). Subset loads now stay **STAGING** (status `partial`); activation catches `ReleaseMatchFailed` → non-fatal `loaded_not_activated` (no crash-loop on an empty required table). +2 regression tests.

Issue #252, opened 2026-07-28
feat(omop): wire mirror into matcher/backfill (Phase-T) + fail-closed + observability


LeoMo42:
Part of epic #246. See ADR 0002 (#245) mechanisms #3/#5/#9 + Invariant.
- **Per-request MatchingReleaseContext:** read `active_release` once at request start; pin **all OMOP-mirror reads** in the run to that `release_id` (wrap the whole eligibility computation, pass the pin explicitly — not a module global). READ COMMITTED + pin + retention window (no REPEATABLE-READ).
- **Phase-T:** trial regimen→component expansion + patient regimen→component expansion run locally against the pinned release.
- **Fail-closed:** no verified bundle → 503 `vocabulary_release_unavailable`; sync failing → serve active bundle until `MAX_MIRROR_AGE`, then 503; never fall back to live API / other release; plain browsing without eligibility may stay up.
- **Observability:** `release_id` in match metadata + `X-Exact-OMOP-Release` header; per-match logging; metrics (active/latest/age/sync-state/mismatch-rejections). Concurrency test: start on R1, activate R2 mid-run → R1 reader stays fully R1.
- **Deferred:** the full cross-artifact/versioned CB trial-projection contract (Conway) — separate ticket tied to when Phase-T ships.

LeoMo42:
**Scope moved in from #251:** the `OmopConcept` title-resolution migration (`_resolve_omop_concepts` → mirror `concept` table) and `ComponentCategoryOmopLookup` release-keying land here, not in #251. Reason: they touch **live matcher code** and need the mirror **live behind the flag** — migrating them while the prod mirror is empty would break response titles / category resolution. So #252 (wiring/cutover) owns:
- wrap the eligibility computation in a per-request `MatchingReleaseContext` (pin `active_release_id()` once);
- swap the Phase-T graph expansion onto `vocab_mirror.traversal.LocalConceptGraph` (drop-in for the retired client, #251/#258);
- migrate `OmopConcept` title resolution to the mirror `concept` table (behind the flag; drop the loader writes once cut over);
- release-key `ComponentCategoryOmopLookup`'s process-LRU (flush on release change) or replace with release-pinned queries;
- fail-closed 503 + `MAX_MIRROR_AGE` + `X-Exact-OMOP-Release` observability.

LeoMo42:
[...]

Issue #251, opened 2026-07-28
refactor(omop): local graph traversal + retire the #234 API+cache surface


LeoMo42:
Part of epic #246. See ADR 0002 (#245) mechanisms #2/#3/#8.
- **Local traversal** service: descendants/ancestors over the mirror (`concept_ancestor`/`concept_relationship`), filtered by the pinned `release_id`.
- **Retire API+cache:** remove `ConceptGraphClient` + `CachedConceptGraphClient` (no runtime caller today) and the `concept_graph` DB cache infra (`cache_config.py` alias, `CONCEPT_GRAPH_CACHE_ALIAS`, the `createcachetable` step) from slice-2b (#240). **No request-time fallback** — a stale/absent mirror fails closed (T7), never a live/other release.
- Migrate `OmopConcept` title resolution (`_resolve_omop_concepts` / matcher) to the mirror `concept` table; drop the loader writes.
- Key `ComponentCategoryOmopLookup`'s process-LRU by `release_id` (flush on release change) or replace with release-pinned queries.

LeoMo42:
Implemented by #258, merged to the `2omop` epic branch. It merged to a non-default branch, so it did not auto-close — closing manually. Reaches prod via the 2omop→dev/main cutover (#223).

Issue #234, opened 2026-07-23
Adopt promop vocabulary SoT via API + cache (ConceptGraphClient, DB-backed, release-pinned, fail-closed)


LeoMo42:
> **⛔ Superseded 2026-07-29 by the vocab mirror — epic #246, ADR 0002 (#245).**
>
> This issue's mechanism — **promop vocabulary SoT via API + client-side cache**
> (`ConceptGraphClient` + DB-backed cache) — is no longer the approach. Per Adam's
> mandate (promop protocol #334, `vocab-consumer-cache-protocol.md`), EXACT instead
> keeps a **release-pinned local vocab mirror** so the *whole* vocabulary is
> consistent client-side (a per-request API + cache cannot guarantee cross-artifact
> whole-vocab consistency — see ADR 0002 for the full "why").
>
> **What replaced it (all merged on `2omop`):**
> - #248 mirror schema, #249 release contract + atomic activation + match gate,
>   #250 vocab sync command (promop #334), #251 local traversal **+ retirement of the
>   `ConceptGraphClient` API + `DatabaseCache` surface this issue built** (T6).
> - #252 / PR #261 wires mirror-backed title resolution + per-request release pin.
>
> The client/cache slices merged under this issue were retired in #251. No further
> work happens here — track the mirror epic at #246. Closing as superseded.

---

Per promop **ADR 0001** (revised 2026-07-24, PR healthkey-ai/promop#279): all data clients
access promop vocabulary + concept-graph data **exclusively via the HTTP API, backed by a
client-side cache**. promop's API **is** the synchronization contract; the cache is a
transient, release-pinned layer — **not** a snapshot/delta replica (that model is rejected)
and **not** a local source of truth.

> **Rescoped 2026-07-24:** the earlier snapshot/delta framing of this issue is superseded.
> Consuming a "versioned release (snapshot + delta)" is no longer the mechanism — it is
> API + cache. See promop#279.

## Scope (EXACT side)

- **`ConceptGraphClient`** — thin client over promop's concept API
  (`GET /api/v1/concepts/graph/`, plus `lookup`/`search` as needed), mirroring the
  `CtomopClient` pattern.
  - Batch (≤200 sources / ≤1000 nodes per source); honor the `truncated` flag as
    **fail-closed**, never silent.
  - **DB-backed cache** (`django.core.cache.backends.db.DatabaseCache`), **not**
[...]

PR #258, opened 2026-07-28
feat(omop): local mirror traversal + retire #234 API+cache (#251)

LeoMo42:
Traversal + retirement slice of the vocab-mirror epic **#246** (ADR 0002). Closes **#251**. **Stacked on #257 (#250)** — base `feat/250-vocab-sync`.

## What
Replace the (never-wired) per-request promop concept-graph API + result cache with **local traversal over the release-pinned mirror**.
- **`vocab_mirror/traversal.py` — `LocalConceptGraph`:** `descendants`/`ancestors`/`expand` (concept_relationship BFS, `max_levels`-bounded, cycle-guarded; `relationship_ids`/`vocabulary_ids`/`concept_class_ids` filters) + `closure_descendants`/`closure_ancestors` (concept_ancestor closure). Every traversal is **release-pinned** (explicit > instance > active) and **fails closed** (`ConceptGraphUnavailable`) — never empty/fail-open, never a live call. Public shape mirrors the retired client → **drop-in** for the Phase-T wiring (#252).
- **Retire the #234 API+cache** (no production caller — never wired): delete `concept_graph_client.py` + `concept_graph_cache.py` + tests; drop the `concept_graph` DB-cache alias (`cache_config.py`), the `CONCEPT_GRAPH_CACHE_ALIAS` setting, and the `createcachetable` deploy step (entrypoint.sh + deploy-staging.yml — that table was the only DB cache).

## Scope note
`OmopConcept` title migration + `ComponentCategoryOmopLookup` release-keying (originally listed here) **move to #252** — they touch live matcher code and need the mirror live behind the flag; migrating now would break titles against an empty prod mirror.

## Review & tests
Two-part self-review (fresh-context agent + codex): retirement verified clean (no dangling refs; deploy safe + valid bash/YAML); BFS traced correct on chain/cycle/diamond + fail-closed. Addressed codex: added `vocabulary_ids`/`concept_class_ids` drop-in params (one-query post-filter), the ADR 0001 note pointer, and filter/diamond/multi-source/pin-precedence/closure tests. Tests: **16** traversal; full suite **1041 passed / 5 skipped** (net of the deleted concept-graph tests); `makemigrations --check` clean.


LeoMo42:
**Holistic whole-stack review fix:** force-pushed a hardening of the traversal pin — it now pins ONE release lazily+cached (no multi-call generation mixing) and verifies the pin is the **ACTIVE** generation (was: trusted any explicit `release_id`, could read STAGING/partial fail-open). Traversal is now active-only by construction. Also narrowed the checksum docstring (verification is follow-up #256). Tests rewritten for active-only/pin-once. Follow-ups filed: #259 (retention reaper), #260 (ops: schedule+secrets+drop orphaned cache table).

Issue #263, opened 2026-07-29
feat(omop): Phase-T — [REDIRECTED by ADR-0004] EXACT-side regimen→component expansion (do not build; promop expands)


LeoMo42:
Split out of #252. The local graph traversal (`vocab_mirror.traversal.LocalConceptGraph`) is built + fail-closed + release-pinned, but NOT on the eligibility path — today's OMOP matching uses materialized trial `omop_*` columns + consumer-supplied patient components (Phase P), no live regimen→component expansion.

Phase-T = the measured decision to flip it on:
- **Measure** how often a regimen concept_id arrives without pre-supplied components (regimen-unresolved frequency) — the ADR's gating metric. If low, Phase-T may not be worth it.
- If pursued: wire `LocalConceptGraph` regimen→component expansion into the matcher/backfill behind the flag; switch its `_pin` to `release_context.active_pinned_release()` (the #252 traversal note) so it honors the request pin; and add the **fail-closed 503** (`vocabulary_release_unavailable` + `MAX_MIRROR_AGE`) that #252 deliberately omitted — because then eligibility WILL depend on the mirror.
- Adjust `LocalConceptGraph` to accept a context-validated pinned release during the retention window (it currently requires strictly ACTIVE; a mid-request activation could otherwise 503).

Depends on the readiness gate/versioning follow-up. Part of epic #246.

LeoMo42:
**Measurement slice** landed in **PR #267** (scope confirmed with the team: measure first, flip deferred).

Instruments the gating metric — regimen present at matching under OMOP with no consumer-supplied components — **once per search** (queryset passes `measure=True`; the per-trial matcher doesn't, to avoid a log flood). Zero behavior change, no dependency on #262/#265.

**Still deferred here** (do once the metric justifies it AND #262/#265 land): wire `LocalConceptGraph` regimen→component into matcher/backfill behind the flag, switch `_pin` → `active_pinned_release()`, add the fail-closed 503 (`MAX_MIRROR_AGE`), and accept a context-validated pinned release during the retention window.

LeoMo42:
**Redirected by [ADR 0004](../blob/2omop/docs/adr/0004-patient-expansion-tell-dont-ask.md) (PR #338) — "tell, don't ask".**

[...]

PR #264, opened 2026-07-29
feat(omop): release-gate the component-category lookup (B′) (#262)

### LINKED ISSUE #262: feat(omop): release-version ComponentCategoryOmopLookup + cross-artifact readiness gate
Split out of #252 (surfaced in the codex design pass). The component→category reverse-map (`ComponentCategoryOmopLookup`) is **eligibility-adjacent** and currently **unversioned**: the table has no `release_id` and its query + process-LRU are unpinned (`component_category_lookup.py`, keyed only by the sorted concept-id tuple). Release-keying the LRU alone is insufficient.

Needs:
- Make the lookup **release-versioned** (add `release_id`, or an activation-time release-match contract for the table) so a stale/wrong lookup can't be paired with a newer mirror release.
- Thread the pinned `release_id` (via `release_context.active_pinned_release()` / explicit param) through `derive_component_and_type_values` and its **three** consumers: matcher, queryset, detail display (`therapy_graph.py`, `querysets/trial.py`, `user_to_trial_attr_matcher.py`).
- Extend the activation **release-match gate** (`vocab_mirror/activation.py`) to the cross-artifacts the ADR §Cross-artifact calls for — the trial `omop_*` projection + this lookup — so a release only activates when they agree (ADR 0002 mechanism #4, currently the gate verifies mirror tables only).

Part of epic #246; do before Phase-T flips the graph onto the eligibility path.

LeoMo42:
Closes #262 (lookup gate). Part of epic #246. Design validated by codex + a fresh-context review.

## What & why
The component→category lookup (`component concept_id → CB category codes`, used for OMOP "type" matching) was eligibility-adjacent but **unversioned** — no `release_id`, an unpinned process-LRU, no cross-artifact consistency check — so a stale/wrong lookup could pair with a newer mirror release.

Its content is CB-authored and promop-release-**independent** (ADR 0001-A: component→type is deliberately NOT OMOP-mapped); its only tie to a release is that its keys are concept_ids that must exist in the pinned release. So per-row `release_id` tagging would be **fake generations**. The codex-validated design (**"B′"**) is: one atomically-published payload + a release-validation **stamp** + an **activation gate**.

## Changes
- **Move** `ComponentCategoryOmopLookup` `trials` → `vocab_mirror` app (writable on `default`, not the read-only `trials` catalog DB — also fixes a latent split-DB bug where the lookup routed to a DB EXACT never migrates). Add `ComponentLookupStamp` (enforced-singleton) recording which release the payload was validated against.
- `sync_component_category_lookup(release_id=None)` writes the stamp in the **same atomic txn** as the payload rebuild; defaults to the active release.
- `activation._component_lookup_matches_release` gate: refuse to activate release R unless the lookup is **stamped for R** (provenance) **and** every lookup concept_id **exists in R's mirror concepts** (coverage). Empty lookup (fresh deploy) never blocks.
- The sync flow rebuilds+stamps the lookup for the release before both `activate_release` call sites (under the sync advisory-lock singleton).
- Data migration (`vocab_mirror 0004`) repopulates the moved table from the M2M on migrate, so type matching is never empty in the window after the move.

Consumers are deliberately **not** release-threaded (B′): one atomically-validated payload, activation blocked unless it matches the active release. Phase-T (#263) threads its own graph pin.

## Verification
- Full suite: **1065 passed, 5 skipped** (fresh DB, exercises the new migrations); `makemigrations --check` clean.
- Two-part review (codex + fresh-context subagent) — no merge-blockers. codex's P1 (empty-table window after the move) fixed via the data migration; one P3 deferred (see below).

## Deferred (tracked follow-ups)
- Trial `omop_*` projection cross-artifact gate — has no release provenance today and lives on the read-only `trials` DB, so it can't be made atomic with this `default`-DB gate. Filed separately.
- Consumer LRU is not release-keyed and `clear_lookup_cache()` only flushes the sync process's LRU → gunicorn workers can serve a stale payload after an out-of-band M2M change without a restart (ADR 0002 guard #8). Low severity (payload release-independent; changes normally deploy-time). Filed separately.


LeoMo42:
Follow-ups filed for the deferred pieces: **#265** (trial `omop_*` projection gate) and **#266** (LRU cross-process flush / ADR #8).

Issue #350, opened 2026-08-09
omop(S4): repoint EXACT reads to projection_snapshot + attestation trials-DB relocation; then flip Gate-1 enforce


LeoMo42:
Trial-side read change for **Design S** (ratified in CB `docs/omop/adr-projection-revocation-trigger.md`; gap #2 write-side). Consumes CB #4695/#4696/#4697. Part of #286 / #283.

**Read repoint (3 seams, all funnel through `THERAPY_MATCH_PROFILE.therapy_types_{required,excluded}`):**
- seams 2+3 (matcher verdict `_match_omop_type_things`, display `therapy_related_things_match_status`): **trivial** — repoint the profile getattr from `self.trial` to the pinned snapshot row.
- seam 1 (queryset prefilter `eligible_for_omop_therapy_types`): repoint the queryset base from `Trial` to `projection_snapshot` (its `jsonb_ops` GIN carries the `__has_any_keys` prefilter), read via the `trials` alias.

**Attestation/registry relocation (shared prereq):** move `ProjectionAttestation` (+ the new registry) off `_DB='default'` onto the CB-owned trials DB routing, and make writes **insert-only** (today `vocab_mirror/attestation.py` writes mutable `update_or_create` on default).

**Read contract (full conjunction):** for the pinned P — `published` ∧ `NOT disabled` ∧ recomputed checksum `== manifest_checksum` ∧ vocab equality on R — one commit-current snapshot on the authoritative primary.

**Then flip** `_PROJECTION_GATE_ENFORCE` + `_PATIENT_RELEASE_GATE_ENFORCE` + `EXACT_OMOP_THERAPY_TYPES` (gap #1 checksum-verify #345 and `type_release_gate` already done; patient-side #337/#344 done). Consider splitting the flip into its own PR after repoint is shadow-verified.

PR #353, opened 2026-08-10
feat(omop): S4a — EXACT read-models for CB projection registry/snapshot (gap #2)

LeoMo42:
Part of #350. Read-side **foundation** for the EXACT read-repoint of ratified **Design S** (CB ADR `docs/omop/adr-projection-revocation-trigger.md`, Gate 1 gap #2). EXACT will read the CB-owned immutable projection snapshot instead of live `Trial` columns — this lands the models + accessor, flag-agnostic; the matcher-seam repoint + full read-contract are **S4b**.

## What
- **`ProjectionRelease` + `ProjectionSnapshot` read-models** (`trials` app) that **mirror CB's schema** (cancerbot #4695) byte-for-byte — same table names (`trials_projectionrelease` / `trials_projectionsnapshot`), columns (incl. shared `TimeStampMixin`), PK, FK `db_column='projection_id'`, unique `(P, trial_code)`, and the two single-column `jsonb` GINs. `TrialsDatabaseRouter` routes them to the CB-owned trials DB (`trials` alias) when `TRIALS_DATABASE_URL` is set, else the local `default` (single-DB / tests). **EXACT never writes these** — CB owns them and enforces insert-only / one-way via DB triggers (EXACT's migration deliberately omits those `RunSQL` guards; `allow_migrate` keeps EXACT from migrating the externally-owned alias).
- **migration 0023** (creates the tables on `default` for single-DB / tests only).
- **`trusted_release_for(vocab_release_id)`** — the registry-state half of the read contract (`published ∧ ¬disabled ∧ vocab equality`, `nulls_last` ordering). The checksum-match half + seam repoint are S4b.

The old `default`-DB `vocab_mirror.ProjectionAttestation` is untouched; S4b switches the activation gate onto this relocated (trials-DB) registry and retires it.

## Tests (10)
model+snapshot roundtrip, unique `(P, trial_code)`, and `trusted_release_for` returning/excluding published/disabled/unpublished/wrong-release/empty/None/non-str/newest/null-published_at-sorts-last. `makemigrations --check` clean.

## Review
`codex review` (fixed a `nulls_last` P2) + fresh-context subagent — **schema mirror verified byte-identical vs CB field-by-field**, router/allow_migrate sound, accessor + coexistence clean.

**Follow-up (P3):** nothing enforces the CB↔EXACT *schema* mirror across repos over time (like the checksum golden-lock #352 does for the hash) — consider a shared schema-fingerprint both repos assert, so a future CB column change can't silently break EXACT's alias read.


LeoMo42:
Closing — gap #2 (immutable snapshot read-repoint) is deferred as an optional future hardening (owner call: within-release drift is tolerated in practice; don't couple it to the PROMOP+EXACT type-matching migration). S1/S2/S3 (CB, merged) + golden-lock #352 stay as dormant, zero-impact foundation; reopen this read-side when gap #2 is revived.

Issue #358, opened 2026-08-10
Federate Find Trials into the PHR portal, and run EXACT on Render


vtrv101:
"Find Trials" is the last disabled item in the portal's My Health menu. EXACT already exposes the component the portal needs; what is missing is an auth path, a way to serve the remote, and somewhere to run — the same shape as the SoC work (soc#262), with one significant difference: PostGIS.

## What already exists

- The remote is built and named: `frontend/vite.remote.config.ts` exposes `./TrialMatches` and `./types` as `exact_remote`, sharing the same React/query singletons as the other remotes so one host can load several.
- `docker-compose.yml` already has both a `db` (postgis/postgis:16-3.5) and an `exact` app service, so a local end-to-end run is close.
- The image already installs `gdal-bin`, `libgdal-dev`, `libgeos-dev` — GeoDjango's native dependencies are handled.
- CORS is env-driven (`CORS_ALLOWED_ORIGINS`, with `CORS_ALLOW_ALL_ORIGINS` for local dev).

## Gaps

**1. EXACT cannot verify portal tokens.** `settings.py:222` lists only `accounts.providers.firebase.FirebaseTokenProvider`. It needs the same portal provider the other services now carry: RS256 verification against the portal's JWKS with introspection fallback. Check whether `Identity.sub` is unique globally or per issuer before choosing whether to namespace the sub — hk-labs and soc differ on this.

**2. The image never builds the remote, and nothing serves it.** The Dockerfile is a single Python stage — no node stage, so `frontend/dist/remote` never exists. There is also no whitenoise configuration at all, unlike soc, which already had `WHITENOISE_ROOT` wired and only needed the build. Both halves have to be added here.

**3. PostGIS on Render.** This is the real unknown and should be settled before any Terraform is written. `settings.py` uses `django.contrib.gis.db.backends.postgis`, so the database must have PostGIS available and the extension created. Verify Render's managed Postgres supports it on the plan we would use; if it does not, this needs a different database story than labs and soc.

**4. An optional second database.** `TRIALS_DATABASE_URL` configures a separate `trials` database. Decide whether staging uses it or runs single-database, because it changes what gets provisioned.

## Plan

**exact**
- Add `accounts/providers/phr.py` mirroring the soc/hk-labs provider, register it, and add `PHR_*` settings. Make `PARTNER_AUTH_PROVIDERS` env-driven so the Render deployment can run portal-only — there is no Firebase there.
[...]

PR #359, opened 2026-08-10
Accept portal-issued tokens, and ship the federation remote in the image

vtrv101:
Part of #358 — the EXACT side of federating Find Trials into the PHR portal.

## Three gaps closed

**1. EXACT could not verify portal tokens.** `settings.py` listed only `FirebaseTokenProvider`. `accounts/providers/phr.py` verifies RS256 against the portal's JWKS (cached, with a refresh floor so a token carrying an unknown `kid` cannot drive an unbounded number of blocking outbound fetches), falling back to RFC 7662 introspection for deployments still on the HS256 dev key. Access tokens only.

Identities are provisioned as `sub="phr:<id>"`, because `Identity.sub` here is `unique=True` **globally** rather than per issuer — a raw portal id could collide with another issuer's subject namespace.

`PARTNER_AUTH_PROVIDERS` is now env-driven, defaulting to both providers so existing deployments are unchanged. Compose sets it to the portal provider alone, which is how a deployment without Firebase should run.

**2. Nothing built the remote.** The Dockerfile had only a Python stage. A node stage now builds the remote into the image, which `CorsWhiteNoiseMiddleware` then serves from the site root at `/remoteEntry.js`.

Rebased since first opening this PR: `main` grew its own whitenoise wiring and that middleware in the meantime, which supersedes the settings changes this branch originally carried, so those were dropped in favour of `main`'s. Only the build stage was still missing — `Dockerfile.gcp` built the remote, the plain `Dockerfile` did not.

**3. No lockfile.** Also resolved on `main` during the rebase; this branch no longer carries one.

## Added for the Render deployment

**`/healthz`** — checks the default database rather than merely proving Django booted, because a probe that stays green through a deploy whose migrations never ran lets the platform route traffic to a service that 500s on every real request. It deliberately skips the `trials` alias: that points at an external host, and an outage there should not pull this service out of rotation when the trial endpoints already degrade on their own.

**`.github/workflows/deploy-render.yml`** — ships `main` to the Render staging service, alongside the existing Cloud Run `deploy-staging`; neither platform is authoritative and this does not touch the GCP path. It hangs off `workflow_run` for the `django` suite because none of this repo's CI workflows are `workflow_call`-able, and rewiring them is a larger change than this deploy justifies. Needs `secrets.RENDER_API_KEY` and `vars.RENDER_EXACT_SERVICE_ID`.

## Local run

Compose points at a portal on the host. CORS is left allow-all locally rather than pinned to a dev port: vite moves to 5174, 5175… whenever the previous port is taken, and a pinned list looks correct right up until it blocks the portal's preflight.

## Verified against the running stack

```
GET /trials/     (no token)     → 401
GET /trials/     (portal token) → 200
GET /countries/  (no token)     → 401
GET /countries/  (portal token) → 200
GET /remoteEntry.js             → 200, Access-Control-Allow-Origin: *
Identity: issuer=healthkey-phr sub=phr:1
```

End to end from the portal, the remote mounts and renders its filter UI.

## Where the trial data comes from

Settled since this PR opened, and worth recording because #358 left it open. The staging deployment reads the **same externally managed trials database** the existing instance does — it is not copied. Copying was considered and rejected: the Render database behind the running service holds only Django scaffolding (`identity` 0 rows, no trials), so there was nothing there to copy, and duplicating the ~271 MB corpus buys isolation at the cost of a snapshot nothing refreshes.

Worth knowing for whoever wires the next environment: `TRIALS_DATABASE_INIT_FROM_BACKUP=1` restores that corpus from a public snapshot, but it **drops the public schema first** — safe against a database of its own, destructive against the shared one. Staging leaves it unset.


LeoMo42 (commented):
## Two-part self-review of this PR (fresh-context review + `codex review`)

Reviewed at `10aaa55`. Merges clean. I test-merged current `main` into the head and ran the PR's own suites there: **158 passed, 1 skipped** (`tests/test_phr_auth.py`, `test_partner_auth_chain.py`, `test_settings_security.py`, `test_health.py`, `test_deploy_render_workflow.py`).

### Landing status

All nine commits of this branch are already ancestors of `origin/dev` — `accounts/providers/phr.py`, `exact/health.py`, `deploy-render.yml` and the tests exist there and nowhere else. They are **not** in `main` or `2omop`. `dev` carries no further changes to any file this PR touches: the whole dev↔head delta is 43 lines in `.env.example` / `docs/setup.md` / `settings.py`, all of it unrelated work (`TRIALS_DB_TOLERATE_MISSING_COLUMNS`, `EXACT_OMOP_THERAPY`, `CACHE force_redis`). So the auth code under review is exactly what is running on `dev`.

### The one thing that changed meaning since this PR was opened

`main` has moved 5 commits. None touches a file here, but one changes the policy this PR implements:

- #329 (23 Aug) removed `push: branches:[main]` from `deploy-staging.yml` — *"EXACT staging is deployed manually, not from CI"*.
- `deploy-render.yml` (added here, written 10 Aug) deploys **every green push to `main`** to Render.
- Its header comment says *"Runs alongside deploy-staging, which ships the same commit to Cloud Run"*. After #329 that is no longer true — in the merged tree `deploy-staging.yml` is `workflow_dispatch:` only.

Not a defect, a decision: either Render is a deliberate exception to the manual-deploy rule, or merging this re-introduces on-push deploys that were deliberately removed. Either way the comment needs updating.

The gating itself is careful and I checked it: `workflow_run.event == 'push'` **and** `conclusion == 'success'` **and** `head_branch == 'main'`, with `workflow_dispatch` separately branch-checked, the commit pinned to `head_sha`, and an explicit note on why `head_branch == 'main'` alone is insufficient for a public repo (a fork PR's run carries the fork's branch name). The `workflows: [backend]` pairing still holds — `django.yml`'s `name:` is unchanged on `main`.

### Measured: which guards the tests actually hold down

Six mutations of the security guards, run against the merged tree:

| mutation | result |
|---|---|
| `require: ["exp"]` dropped from `jwt.decode` | **1 failed** — caught |
| `PHR_ALLOW_INTROSPECTION` gate removed | **1 failed** — caught |
| `active is not True` → `not active` | **7 failed** — caught |
| empty `sub` accepted | **1 failed** — caught |
| `algorithms=["RS256"]` → `["RS256", "HS256"]` | 153 passed — not caught |
| `issuer=settings.PHR_ISSUER` removed from `jwt.decode` | 153 passed — not caught |

The two survivors are **not holes**. Only `alg == "RS256"` reaches `_verify_jwks` (`phr.py:257`), and only `iss == PHR_ISSUER` reaches the provider at all (`can_handle`, `phr.py:238`) — both mutated lines are the second layer, whose first layer is elsewhere and is tested. Worth a cheap test each so the layering is not silently lost later.

### Findings

**P1 — `is_active` is never enforced for portal identities.** `PartnerAuthentication.authenticate` returns the `Identity` directly (`accounts/authentication.py:74-76`) and `AbstractBaseUser.is_authenticated` is unconditionally `True`, so `IsAuthenticated` passes; `accounts/admin.py` exposes an `is_active` toggle that does nothing here. DRF's own `TokenAuthentication` does check it (`rest_framework/authentication.py:205`), so revocation behaves differently per credential type. *Scenario:* an operator disables a compromised portal identity in `/admin/`; its still-valid portal token keeps working until the portal stops minting. `accounts/authentication.py` is **not in this diff** — pre-existing — but this PR is what routes portal users through it. `ServiceTokenAuthentication` has the same gap.

**P1 — `aud` is not validated.** `options={"verify_aud": False, ...}` (`phr.py:309`); there is no `PHR_AUDIENCE` setting, and `test_phr_auth.py:388-394` asserts the *absence* of the control and documents it as accepted risk. *Scenario:* a token the portal issues for a different relying party — or one lifted from that party's logs — authenticates here with full access. The only thing holding it is the portal currently having one audience, an external invariant nothing in this repo pins. If it stays accepted, worth recording the condition under which it must be closed.

**P2**
- `PHR_JWKS_URL` scheme is not constrained; `.env.example` and `docker-compose.yml` ship `http://`. A plaintext JWKS fetch lets an on-path attacker substitute a signing key — a complete bypass. Operator-controlled, but nothing warns.
- The verified-token cache (`AUTH_TOKEN_CACHE_TTL`, 60s, keyed on `sha256(token)[:32]`) stores no `exp`, so a token expiring inside the window keeps working. Pre-existing; the key is not attacker-chosen.
- `FirebaseTokenProvider.can_handle` does `iss.startswith(...)` (`firebase.py:26`) outside the caller's `try`, so a non-string `iss` is an unauthenticated 500. Out of the changed files, but this PR makes Firebase→PHR the default chain and the new chain tests do not cover it.
- JWKS revocation latency is up to `PHR_JWKS_CACHE_TTL` (3600s); only an *unknown* kid forces an early refresh.
- `/healthz` inherits `AnonRateThrottle` (60/min) — a shared-egress or high-frequency probe can 429 and pull the service out of rotation.
- `docker-compose.yml:52` uses `host.docker.internal`, which is not defined on native Linux, so both JWKS and introspection fail there (`extra_hosts: host.docker.internal:host-gateway` fixes it). — from `codex review`.

**Test gaps**: `test_settings_security.py::test_introspection_default_rule_fails_closed` defines `enabled()` inside the test and asserts against its own function — it would pass verbatim if the setting were deleted from `settings.py`; the two regex tests above it do bind to the source, but no test imports settings under a cleared `ENVIRONMENT` and asserts the resolved boolean. No coverage for the `issuer=` pin (mutation above), the deactivated identity, the 60s auth-cache window, or the JWKS URL scheme.

**Retracted during verification:** a reported "non-string `kid` → uncaught `TypeError` → one warning log per anonymous request" is **not real**. PyJWT rejects a non-string `kid` in `get_unverified_header` (`_validate_kid`, api_jws.py:432) before it can reach the cache; verified by hand — `kid={}` and `kid=["a"]` both return `None` with no exception.

### Verified sound — please don't re-litigate

Alg confusion, `alg=none` and HS-signed-with-the-RSA-public-key are closed by the explicit dispatch allowlist plus `algorithms=["RS256"]`, and the tests prove it by asserting empty *call lists*, not just `is None`. `iss` is pinned via `issuer=`, `exp` is `require`d (closing PyJWT's no-op-when-absent hole), `nbf` is checked. The unverified payload is used only to reject or route — the one place it supplies a value is the introspection subject fallback, which is gated off outside DEBUG/local. Introspection is fail-closed on unset `ENVIRONMENT` (read via `os.environ.get`, not the module-level default) and no longer derives its URL from `PHR_BASE_URL`. Outbound work is bounded by the JWKS refresh floor and the fixed-window call budget, both with tests that fail when the guard is removed; httpx defaults keep TLS verification on and redirects off, with a 5s timeout and no retries. `PHR_INTROSPECT_AUTH` is split with `partition` and a half-credential is refused rather than sent. `active` is checked with `is not True`, so `"false"` cannot read truthy. The JWKS cache cannot be grown or poisoned by a token, a broken document never evicts good keys, and rotation is picked up. Provisioning keys only on `(issuer, sub)` — no email linking, so no takeover of an existing account — with an unusable password and `is_staff`/`is_superuser` untouched. No chain downgrade: `TokenAuthentication` needs the `Token ` keyword, `SessionAuthentication` ignores the header, and a swallowed provider exception yields 401, not anonymous. `request.auth` is read nowhere in `trials/`, `accounts/` or `exact/`, so the merged claims feed no authorization decision. No token, header or payload is logged.

### Suggested disposition

1. Decide the Render auto-deploy question and fix the stale `deploy-staging` comment — it is the only thing whose meaning changed while this PR sat open.
2. File the `is_active` gap and the `aud` decision as issues; neither blocks this PR, both outlive it.
3. The two uncaught mutations are cheap tests worth adding so the second layer is not lost silently.

🤖 Two-part self-review: fresh-context Claude review + `codex review`. Mutation and merge-tree results are measured, not inferred; one reported finding was retracted after failing to reproduce.

LeoMo42 (approved):
Approving on the auth work — with one condition to settle before this lands.

**The condition.** `deploy-render.yml` deploys every green push to `main`, while `main` itself moved to manual-only deploys in #329 ("EXACT staging is deployed manually, not from CI"), and this workflow's header still describes `deploy-staging` as shipping the same commit to Cloud Run, which it no longer does. Either Render is a deliberate exception to that rule — say so in the comment — or the trigger should be `workflow_dispatch:` only, matching `deploy-staging`. My review comment above has the details. This is a decision, not a defect; I just do not want the two deploy policies to diverge silently.

**What I am approving.** The RS256 path holds up: `alg` pinned by an explicit dispatch allowlist plus `algorithms=["RS256"]`, `iss` pinned, `exp` required, the unverified payload used only to reject or route, introspection fail-closed on an unset `ENVIRONMENT` and no longer derived from `PHR_BASE_URL`, outbound work bounded by the refresh floor and the call budget, provisioning keyed only on `(issuer, sub)`. Four mutations of the load-bearing guards (`require: ["exp"]`, the `PHR_ALLOW_INTROSPECTION` gate, `active is not True`, the empty-`sub` guard) each fail the suite, so those properties are genuinely held down by tests, not merely asserted in comments. 158 passed / 1 skipped with current `main` merged in.

**Not blocking, but worth issues:** `is_active` is never enforced for portal identities (`accounts/authentication.py:74-76`, pre-existing and outside this diff, but this PR makes that path live), and `aud` is not validated — accepted deliberately today, so it is worth recording the condition under which it must be closed. Both outlive this PR.

LeoMo42 (approved):

LeoMo42:
Ran the pre-merge delta check: `main` has moved 5 commits since this branch
forked, **none of them touching any of the 15 files here**, and `main` has not
moved at all since the approval. Nothing to re-review on that front.

One thing did change underneath this PR, though, and it is live rather than
theoretical.

## This would reintroduce automatic Render deploys on `main`

`#516` (merged to `dev` today) makes `deploy-render.yml` **manual only**:

```yaml
# dev, after #516
on:
  workflow_dispatch:
```

This PR adds the original version to `main`:

```yaml
# what #359 would add
on:
  workflow_run:
    workflows: [backend]
    types: [completed]
  workflow_dispatch:
```

That is not latent. GitHub resolves `workflow_run` from the **default branch's**
copy, the default branch is still `main`, and `main` has no copy today — which is
exactly why #516's description says the trigger "has never fired". Merging this
gives `main` a copy, and every green `backend` push to `main` starts a Render
deploy from that moment.

It also runs against the decision in #329 / `d3d3cb3`, which took
`deploy-staging` down to manual dispatch on 2026-08-23 — twelve days *after*
`deploy-render.yml` was written here (2026-08-11). The file's own header still
describes the older world:

> Runs alongside deploy-staging, which ships the same commit to Cloud Run

`deploy-staging` no longer runs on its own.

## Not a safety hole

Worth saying plainly, because the guard here is careful: the job requires
`workflow_run.event == 'push'` (so fork PRs cannot reach it), `conclusion ==
'success'`, `head_branch == 'main'`, pins `commitId` to the commit the suite
ran against, and Render's own auto-deploy is off. This is a **policy**
divergence, not a vulnerability.

## Suggested resolution

Take `dev`'s post-#516 version of `deploy-render.yml` (and
`tests/test_deploy_render_workflow.py`, which #516 also updated) onto this
branch, so `main` and `dev` land byte-identical and staging stays manual
everywhere. Happy to prepare that commit if you want it.

The alternative is deciding that Render staging *should* auto-deploy while Cloud
Run staging does not — a defensible call, but one worth making explicitly rather
than inheriting from a file written before #329.

Issue #360, opened 2026-08-10
Trials corpus schema is behind the models — /trials/ 500s on every deployment


vtrv101:
`GET /trials/` returns 500 against the shared trials corpus: the database is behind the models.

```
django.db.utils.ProgrammingError: column trials_trial.omop_intervention_concept_ids does not exist
```

## Not staging-specific

`omop_intervention_concept_ids` is added by `trials/migrations/0008`, and **`dev` — the branch `exact-2` deploys — declares it too**, along with `TrialUniverse` and `RawDataItem`. So any current deployment reading this corpus hits the same failure on `/trials/`. It surfaced now only because the Render staging service is the first one exercising that endpoint with a portal token.

The public snapshot (`cancerbot-public/exact/trials/latest.sql.gz`) carries the **same stale schema**, so restoring a fresh database from it reproduces the bug rather than avoiding it.

## The gap

Diffed the models against `information_schema` on `ne_bc_trials`:

| | missing |
|---|---|
| `trials_trial` columns | `omop_intervention_concept_ids`, `lesion_size_mcl_min`, `lesion_size_mcl_max` |
| tables | `trials_rawdataitem`, `trials_trialuniverse`, `trials_trialuniverseentry` |

The corpus has **no `django_migrations` table** — it is a data-only dump, not a Django-managed schema, and `TrialsDatabaseRouter.allow_migrate` returns `False` for the `trials` app by design. So `migrate` will never close this gap; the DDL has to be applied by hand.

Separately, 10 indexes the models declare are absent from the corpus. Unrelated to this failure and left alone.

## Fix

Additive DDL, generated from the models — no existing column or row is touched, and it is reversible:

```sql
BEGIN;
  CREATE TABLE "trials_rawdataitem" (...);
  ALTER TABLE "trials_trial" ADD COLUMN "omop_intervention_concept_ids" jsonb DEFAULT '[]'::jsonb NOT NULL;
  ALTER TABLE "trials_trial" ADD COLUMN "lesion_size_mcl_min" double precision NULL;
  ALTER TABLE "trials_trial" ADD COLUMN "lesion_size_mcl_max" double precision NULL;
  CREATE TABLE "trials_trialuniverse" (...);
  CREATE TABLE "trials_trialuniverseentry" (...);
COMMIT;
```

[...]

PR #361, opened 2026-08-10
Let the trial endpoints work against a corpus behind the models

vtrv101:
Stacked on #359. Stopgap for #360.

`GET /trials/` 500s against the shared trials corpus because the corpus is behind the models:

```
django.db.utils.ProgrammingError: column trials_trial.omop_intervention_concept_ids does not exist
```

The corpus is data-only — no `django_migrations` table — and `TrialsDatabaseRouter.allow_migrate` refuses to migrate the `trials` app into it, by design, because the schema belongs to someone else. So the models drift ahead of it and nothing catches the drift until an endpoint fails.

## What this does

With `TRIALS_DB_TOLERATE_MISSING_COLUMNS` set, every manager in the app is wrapped so its querysets defer the columns the database lacks. Off by default — a deployment whose database matches the models is untouched.

Applied at the manager level rather than per view because per view did not hold: deferring only in `TrialsViewSet` fixed `/trials/` and left `/form-settings/` failing on `trials_therapy.omop_concept_id`.

## Scope, measured

Against `main` the corpus is missing 3 columns and 3 tables, and with this change a deployed staging service answers every endpoint the portal uses:

```
/healthz 200   /remoteEntry.js 200   /trials/ 200 (19 trials)
/countries/ 200   /form-settings/ 200
/trials/ without a token 401
```

`/trials/<id>/` still 500s, but not for a schema reason and not one this PR
introduces: `retrieve` always runs the attribute matcher, and
`user_to_trial_attr_matcher` dereferences `patient_info.prior_therapy` with no
`None` guard, so a detail request carrying no patient context raises
`AttributeError`. Filed separately as #362; none of the files involved are
touched here.

**Known limit, recorded so nobody assumes more of this than it gives.** Against `dev` the same corpus is missing 19 columns and 5 tables, and this approach is *not* sufficient there:

- `select_related('therapy')` in `value_options.py` composes columns from the related model's field list, so no manager-level deferral reaches it.
- `therapy_match_profile`, `therapy_concept_mapper` and `component_category_lookup` **filter** on `omop_concept_id` / `omop_therapies_*`. Deferral only removes columns from `SELECT`; a filter on a column that does not exist can never succeed.

The second point is the one that matters: on `dev` the corpus is missing the OMOP layer the therapy matching is computed from, and forcing those queries through would return matches derived from absent data — worse than a 500. #360 is the fix; this only buys room on `main`.

## Deliberately narrow

- **A missing table defers nothing.** That is a relation the corpus has never held; serving empty results would hide a real gap.
- **Deferred columns still raise if read.** Django reloads a deferred attribute on access, so a filter over data the corpus lacks fails loudly instead of matching on nothing.
- **Introspection failure defers nothing.** If we cannot tell, we do not guess.
- **Related managers are not wrapped** — one more reason this is a stopgap and not an abstraction to build on.

## Verification

Against the real corpus (read-only credentials): undeferred fetch raises `ProgrammingError`, deferred fetch returns rows. 9 unit tests; `tests/api` and `tests/querysets` green at 136 passed.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -22,6 +22,19 @@ PATIENT_DATABASE_URL=postgresql://user:password@host:5432/dbname
 REDIS_URL=redis://localhost:6379
 ENVIRONMENT=local
 
+# ── Portal (PHR) identity ────────────────────────────────────────────────
+# Verifies tokens a signed-in patient carries in from the portal. Leave
+# PHR_BASE_URL unset to disable portal auth entirely (it then fails closed).
+# PARTNER_AUTH_PROVIDERS=accounts.providers.phr.PhrTokenProvider
+# PHR_BASE_URL=http://localhost:8000
+# PHR_ISSUER=healthkey-phr
+#
+# HS256 introspection fallback. On by default only for ENVIRONMENT=local or
+# DEBUG, and it must stay OFF when deployed: it delegates the signature check
+# to the portal, so a portal that introspects without fully verifying
+# signatures would let a forged token through. See docs/setup.md.
+# PHR_ALLOW_INTROSPECTION=false
+
 # PostGIS (if not in system PATH)
 # macOS (Homebrew): GDAL_LIBRARY_PATH=/opt/homebrew/lib/libgdal.dylib
 # macOS (Homebrew): GEOS_LIBRARY_PATH=/opt/homebrew/lib/libgeos_c.dylib
--- a/.github/workflows/deploy-render.yml
+++ b/.github/workflows/deploy-render.yml
@@ -4,15 +4,16 @@ name: Deploy to Render
 # which ships the same commit to Cloud Run: exact is deployed to both, and
 # neither platform is authoritative.
 #
-# Gated on the django workflow rather than calling a reusable one — none of
+# Gated on the backend suite rather than calling a reusable workflow — none of
 # this repo's CI workflows are workflow_call-able, and rewiring them to be so
 # is a bigger change than this deploy warrants. workflow_run fires once that
 # suite finishes, and the conclusion check below is what keeps a red build
 # from shipping.
 #
 # Render's own auto-deploy is off for this service (Terraform sets
-# auto_deploy = false) so the platform cannot race this workflow and ship a
-# commit that never ran the test suite.
+# auto_deploy = false), and every deploy this workflow requests is pinned to
+# the commit that passed the suite — so neither the platform nor a push
+# landing mid-run can ship a revision the tests never saw.
 #
 # Needs:
 #   secrets.RENDER_API_KEY          — Render API key
@@ -20,7 +21,12 @@ name: Deploy to Render
 
 on:
   workflow_run:
-    workflows: [django]
+    # This matches the triggering workflow's `name:`, NOT its filename — the
+    # test suite lives in django.yml but is named `backend`. Naming the file
+    # here silently never matches: workflow_run simply does not fire, every
+    # `if:` below becomes unreachable, and nothing reports an error.
+    # tests/test_deploy_render_workflow.py guards the pairing.
+    workflows: [backend]
     types: [completed]
   workflow_dispatch:
 
@@ -35,9 +41,25 @@ jobs:
     # workflow_run fires for every branch and for failed runs too, so the
     # branch and conclusion both have to be checked here. workflow_dispatch
     # carries neither, hence the event fallback.
+    # Three things have to hold, and `head_branch == 'main'` is not one of them
+    # on its own. This repo is public and django.yml runs on bare
+    # `pull_request:`, so a fork PR gets a `backend` run whose `head_branch` is
+    # the branch name *in the fork* — `main`, for anyone who pushes to their
+    # fork's default branch before opening the PR. That run's completion would
+    # satisfy a branch-and-conclusion check, and this job would then execute in
+    # the base-repo context, with RENDER_API_KEY, pinning `commitId` to an
+    # unreviewed fork commit. Requiring the upstream event to be a `push`
+    # excludes every PR run, fork or not.
+    #
+    # A manual dispatch is branch-checked too. It used to be harmless without
+    # one: the request carried no commit, so Render built the tracked branch
+    # whatever ref the dispatch came from. Now that the commit is pinned, an
+    # unguarded dispatch would ship the dispatched branch's untested tip.
     if: >-
-      github.event_name == 'workflow_dispatch' ||
-      (github.event.workflow_run.conclusion == 'success' &&
+      (github.event_name == 'workflow_dispatch' &&
+       github.ref == 'refs/heads/main') ||
+      (github.event.workflow_run.event == 'push' &&
+       github.event.workflow_run.conclusion == 'success' &&
        github.event.workflow_run.head_branch == 'main')
     steps:
       - name: Trigger Render deploy
@@ -45,18 +67,38 @@ jobs:
         env:
           RENDER_API_KEY: ${{ secrets.RENDER_API_KEY }}
           SERVICE_ID: ${{ vars.RENDER_EXACT_SERVICE_ID }}
+          # The revision the suite actually ran against. Via env, never
+          # interpolated into the script below. workflow_dispatch carries no
+          # workflow_run payload, so it falls back to the dispatched ref's tip.
+          COMMIT_ID: ${{ github.event.workflow_run.head_sha || github.sha }}
         run: |
           set -euo pipefail
 
           if [ -z "${SERVICE_ID:-}" ]; then
             echo "vars.RENDER_EXACT_SERVICE_ID is not set"; exit 1
           fi
+          if [ -z "${COMMIT_ID:-}" ]; then
+            echo "could not resolve the commit to deploy"; exit 1
+          fi
+
+          # `commitId`, not `{}`: without it Render builds whatever the tracked
+          # branch points at when it receives the call, so a push landing while
+          # the suite was still running would ship a revision that never ran
+          # the tests — defeating the gate this workflow exists to be.
+          # On its own line, not inline in the curl arguments: `set -e` ignores a
+          # non-zero exit from a command substitution used as an argument, so a
+          # broken generator would post `-d ""` — the branch-tip deploy this is
+          # here to prevent — instead of failing at the real cause.
+          BODY_JSON=$(python3 -c "
+          import json, os
+          print(json.dumps({'commitId': os.environ['COMMIT_ID']}))
+          ")
 
           RESPONSE=$(curl -s -w '\n%{http_code}' -X POST \
             -H "Authorization: Bearer $RENDER_API_KEY" \
             -H "Content-Type: application/json" \
             "https://api.render.com/v1/services/$SERVICE_ID/deploys" \
-            -d '{}')
+            -d "$BODY_JSON")
           CODE=${RESPONSE##*$'\n'}
           BODY=${RESPONSE%$'\n'*}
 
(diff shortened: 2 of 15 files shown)


LeoMo42:
Cross-posting from #365 so it is on the mechanism PR too.

The same failure (#360) now also has a second fix in flight, from the opposite direction: make the models a strict **subset** of the catalog. Branches `harvard/catalog-schema-fix` (base `main`) and `align/cb-catalog-schema` (base `2omop`). Among other things they delete `Trial.omop_intervention_concept_ids` — the exact column in the traceback quoted in this PR's description — because CB has never had it and never populated it, so `?therapy_id=` was matching zero trials everywhere (context in #379).

Verified against a 2026-08-14 CB production dump and the Harvard-pruned variant of it: `MISSING TABLES (0)` / `MISSING COLUMNS (0)`, all endpoints 200 with no DDL applied to the catalog.

Two spots here to adjust when that lands (details and suggested fix in my comment on #365):

- `trials/db_compat.py:33-36` — the "the live case is `?therapy_id=`" paragraph goes stale, that path stops existing.
- the `ABSENT` fixture in `tests/test_db_compat.py` is that same column, so the suite breaks regardless of merge order.

The mechanism itself keeps its value: it covers the case the subset rule cannot — a catalog that lags *during* a CB rollout, which is live right now for the `2omop` line waiting on CB's `omop_therapy_types_*`. What neither approach covers is a catalog missing a whole **table**: a deployment was found running a snapshot without `trials_trialpurpose`, where this tolerance saves list/search/match without a `trialPurpose` filter but `/form-settings/` and trial detail still 500 (`TrialAttributes.__init__` builds `ValueOptions().all_options()` at `trial_attributes.py:39`, which queries `TrialPurpose`). That one only a fresh catalog transfer fixes.

LeoMo42 (commented):
## Two-part self-review of this PR (fresh-context review + `codex review`)

Reviewed at `0061f10` against base `feat/phr-federation`. Merges clean; base has moved 3 commits, none of them in a file this PR touches. `pytest tests/test_db_compat.py --create-db` → **10 passed** (with `--reuse-db` one test fails on a test database built from another branch — not a defect here).

### First: most of this has already landed, in a later form

All three commits of this branch (`1c13cfd`, `2edbe62`, `0061f10`) are ancestors of `origin/dev`. They arrived there via #414, which merged `fix/legacy-trials-schema-review` — i.e. the **#361 + #365 stack** — into `feat/mf-bridge-trialmatches` and then into `dev`. They are **not** in `main`, `2omop`, or this PR's own base `feat/phr-federation`.

What is on `dev` is not the code in this PR. It is #365's version: `trials/db_compat.py` 122 → ~540 lines, `tests/test_db_compat.py` 230 → 919 lines. `dev` and `fix/legacy-trials-schema-review` are byte-identical on both files.

**So the review question is not "is this good", it is "which version reaches `main`".** Merging #361 alone into the `main` line ships the un-hardened version, including the P0 below.

### Measured: the tests here do not hold the mechanism down

Two mutations, run on both versions (`--create-db`):

| mutation | this PR | `dev` / #365 |
|---|---|---|
| `defer_missing_columns` reduced to `return queryset` — the whole mechanism gutted | **10 passed** (green) | **5 failed** (caught) |
| the `TRIALS_DB_TOLERATE_MISSING_COLUMNS` guard deleted from `missing_columns` — the off-switch removed | **10 passed** (green) | **5 failed** (caught) |

On `dev`, `TestAgainstARealDriftedSchema` drops a real column inside the test transaction and exercises real introspection and a real deferral; `TestOffByDefault` replaces the `property`-object monkeypatch. Both mutations die there.

### Findings, and where each stands on `dev`

| sev | finding | `dev` |
|---|---|---|
| P0 | The tolerance is not restricted to the `trials` alias. `TrialsDatabaseRouter` falls back to `default` when `TRIALS_DATABASE_URL` is unset, and `defer_missing_columns` passes `queryset.db` straight through (`db_compat.py:83` → `:44`), so with the flag on, an **un-applied migration on our own database** becomes a service quietly answering with fewer fields instead of failing | **fixed** — `corpus_alias()`, a `django_migrations` backstop, and same-database detection |
| P0 | No test exercises the deferral (measured above) | **fixed** |
| P1 | `test_disabled_by_default` monkeypatches `connections` with a `property` object; `connections[using]` raises `TypeError`, the broad `except Exception` swallows it, and the test passes with the guard deleted | **fixed** |
| P1 | `db_manager(alias)` silently loses the alias and hints: `types.MethodType` binds the wrapper to the *original* manager, while `db_manager` returns a `copy.copy` with `_db` set. Verified: patched → `_db is None`, hints `{}` | **fixed** — deferring *subclass* instead of an instance method |
| P1 | `apps.clear_cache()` uninstalls the feature, silently and permanently: `_meta.managers` is a `cached_property` returning copies, and `install()` never touches `local_managers`. `ready()` does not re-run | **fixed** (`test_survives_a_registry_cache_clear`) |
| P2 | `missing_columns(model, None)` for an unsaved instance (`trial._state.db`) → `ConnectionDoesNotExist`, swallowed, full traceback logged on **every** `details()` call | **fixed** |
| P2 | Introspection failure is not cached, success is cached forever → one connection attempt plus one traceback per queryset built, for the length of an outage | **fixed** (`_failed_at`, 60s retry interval) |
| P2 | `install()`'s docstring cites `trials_therapy.omop_concept_id` as the motivating case; `Therapy` declares no such field, so `missing_columns` could never report it | **fixed** |
| **P0** | **Only `details()` is guarded.** `TrialSerializer.to_representation` → `instance.attrs_to_fill_in()` → `UserToTrialAttrMatcher`, which does bare `getattr(self.trial, ...)` with no missing-column guard. A drifted column that the matcher reads still 500s `GET /trials/` and `POST /trials/<id>/match/` — deferral makes the read re-query. Today this holds only because the real corpus happens to lack three OMOP columns the matcher does not read | **still open** |
| P0 | `select_related` is not covered — the related model's columns are composed from its own field list. `ValueOptions` does `select_related('therapy')`, so `/form-settings/` — named in `install()`'s docstring as an endpoint this fixes — is **not** covered | open, documented on `dev` |
| P1 | `_base_manager` is never patched (auto-created, absent from `_meta.managers`). Forward-FK dereference goes through it — `trialType = StringRelatedField(source='trial_type')` on every list row | open, documented + test on `dev` |
| P2 | Swallowing the introspection error inside a caller's transaction leaves it aborted; latent in prod (no `ATOMIC_REQUESTS`), but every `django_db` test runs in atomic, so a flaky introspection would surface as unrelated cascading failures | open, deliberate on `dev`; `except` narrowed to `(Error, ConnectionDoesNotExist)` |
| P2 | `values()` / `values_list()` with no arguments call `clear_deferred_loading()` and re-select everything | open, documented on `dev` |
| **P2** | **`details()` drops the criterion row while the matcher still scores that attribute** — the eligibility table and the score disagree about which criteria were evaluated. Rendering `unknown` is truthful; omitting the row is not | **open everywhere** |

Sound, and worth not re-litigating: `information_schema.columns` does cover views; empty `present` → defer nothing and `except → ()` both degrade to today's 500 rather than to silently wrong data; `_original=original` avoids the late-binding closure bug; `deconstruct()` and `use_in_migrations` are unaffected; with the flag off `details()` is byte-identical and `missing_columns` returns before touching `connections`.

### Suggested disposition

1. **Do not merge #361 into the `main` line on its own** — that ships the version with the `default`-database P0. Either land it as a stack with #365, or close both and port `dev`'s version in one PR.
2. File the matcher/serializer gap (first P0 above). The PR's claim that the trial endpoints work is currently a property of *which* columns happen to have drifted, not of the mechanism.
3. Decide `skip` vs `unknown` for the dropped criterion row.
4. `ABSENT = 'omop_intervention_concept_ids'` is still the fixture on `dev` too. `harvard/catalog-schema-fix` and `align/cb-catalog-schema` are merged nowhere (checked), so the collision flagged earlier is dormant — but it will fire in either merge order when they land.

🤖 Two-part self-review: fresh-context Claude review + `codex review`. Mutation results are measured, not inferred.

LeoMo42 (changes requested):

LeoMo42 (changes requested):
Requesting changes — on the version this PR would land, not on the idea, which I think is worth having.

**The blocker.** As it stands, this branch defers on whatever alias the queryset resolves to. `TrialsDatabaseRouter` falls back to `default` for trials models when `TRIALS_DATABASE_URL` is unset, and `defer_missing_columns` passes `queryset.db` straight through, so with the flag on in a single-database deployment an **un-applied migration on our own database** stops being a loud `ProgrammingError` and becomes a service quietly answering with fewer fields. That is a worse failure than the 500 this exists to prevent, and nothing in the branch guards against it.

Two things make it more than theoretical. The flag's own off-switch is untested — deleting the settings guard in `missing_columns` leaves the suite green, because `test_disabled_by_default` monkeypatches `connections` with a `property` object whose `TypeError` the broad `except Exception` swallows. And the mechanism itself is untested — reducing `defer_missing_columns` to `return queryset` also leaves the suite green. Both measured, on this head, with `--create-db`.

**This is already solved.** #365 fixes exactly these: the tolerance is restricted to the corpus alias, backstopped by refusing any database carrying a `django_migrations` table, and `TestAgainstARealDriftedSchema` drops a real column and exercises a real deferral. Both mutations above fail there (5 tests each). That version is also what is on `dev` today — all three commits of this branch reached `dev` via #414, which merged the #361+#365 stack.

**So either resolution works for me:**

1. Land this only as a stack with #365 — never #361 alone into the `main` line; or
2. close both and port `dev`'s version in one PR against `feat/phr-federation`.

I'd lean to (2): it is the version that has been running, and it avoids a window where `main` carries the un-hardened shape.

**Not blocking this PR, but they should not get lost** — both are open on `dev` too, so they need issues rather than fixes here: only `details()` is guarded, while the serializer path reaches the matcher's bare `getattr(self.trial, ...)`, so a drifted column the matcher reads still 500s `GET /trials/` and `POST /trials/<id>/match/`; and `details()` drops the criterion row while the matcher still scores that attribute, so the eligibility table and the score can disagree about what was evaluated. Details in my review comment above.

LeoMo42:
## The one question that decides both PRs, measured

#365's `django_migrations` backstop is the first guard here that does not depend on configuration text, and a fresh-context review could not defeat it — every miss it found fails *closed*. That is good engineering. It is also why I went and checked what our corpus actually looks like.

**Every trials corpus on this machine carries `django_migrations`:**

```
exact_mm             django_migrations present (41 rows)   3114 trials
exact_multi          present (51)                         18462
exact_omoptest       present (51)                         13067
exact_shadow_221     present (51)                         13067
exact_legacy_corpus  present (31)                             1
```

`exact_legacy_corpus` is the byte-copy of the actual drifted corpus this work exists for.

End to end, with the flag ON and `TRIALS_DATABASE_URL` pointed at that copy:

```
флаг включён:      True
алиас корпуса:     trials
отложено колонок:  ()
```

and the log line:

> `'trials' has a django_migrations table, so it is a database this project migrates rather than an externally owned corpus. Deferring nothing… (A corpus that carries a stray django_migrations cannot use this tolerance.)`

So on the database this was written for, **#361 + #365 defer nothing.** The corpus is a dump of the upstream CancerBot Django database, and a full dump carries that table — which is the normal way to produce it.

## What does work, measured on the same database

```
really missing:                             ['omop_intervention_concept_ids']
Trial.objects.all()                          -> ProgrammingError
Trial.objects.defer('omop_intervention_…')   -> OK
```

One column, read nowhere, closed by one `.defer()` on the trials queryset. No flag, no introspection, no manager subclassing, no guards to keep ahead of configuration text.

**Correction to something I said earlier in this review thread:** I proposed hoisting the unconditional `getattr` in `trial_attributes.py`. That does **not** work — `omop_intervention_concept_ids` is a plain field in the `SELECT` of every trial fetch, so the query dies before any attribute is read. The conclusion survived; the mechanism I named did not. Credit to the review that measured it rather than taking my word.

## Recommendation

Close both, and land the `.defer()` against #360. Keep `trials/db_compat.py` on the shelf for the day a corpus drifts by more than one unread column — the two-checks-plus-backstop design is the right shape *if* runtime schema tolerance is ever genuinely needed.

Before closing, the two P1s the review of #365 found are fixed regardless (incoming PR) — the branch should not sit with a known wrong-results bug even if it is then shelved:

- `Deferring.using()` dropped the relation filter on related managers. Measured on the corpus: `trial.locationtrial_set.count()` = 6, `.using('trials').count()` = **43021**, the whole table, with no exception and no log line. A regression #365 introduced by moving from per-instance patching to a subclass, which put `Deferring` in the middle of every related manager's MRO.
- Deleting `install(self)` from `TrialsConfig.ready` left all 922 tests green — the only wire between the module and Django, untested.

LeoMo42:
Closing. The measurement that settles it is in the comment above: **every trials corpus carries `django_migrations`, including a byte-copy of the real drifted one**, so with the flag on and the alias correctly configured this defers zero columns and logs a refusal. #361 + #365 are inert against the database they were written for.

I cross-checked the decision with codex rather than only my own reading. It agrees on closing, and corrected me on three things worth recording here:

1. **"The column is read nowhere as an attribute" was too strong** — the generic detail-field walker reads it indirectly. That is why the detail page keeps 500ing whatever the SELECT list says.
2. **"Keep `db_compat.py` on the shelf" is questionable.** Its limits are *structural*, not unfinished polish: if drift ever touches filters, relations, reflective readers or related models it fails unpredictably. Worth preserving the design discussion — not this implementation as the assumed future answer.
3. **#360 must stay open.** A partial fix that closes the issue hides the remaining failures and makes the workaround permanent by accident.

## The two P1s are fixed regardless — `b20c5be`, branch `fix/365-related-manager-using`

Pushed rather than PR'd, since this branch is closing. The work should not be lost and the branch should not sit with a known wrong-results bug.

- **`Deferring.using()` dropped the relation filter.** Moving from per-instance patching to a subclass put `Deferring` in the MIDDLE of every related manager's MRO, so `super(Deferring, self).get_queryset()` skipped `RelatedManager.get_queryset` — where `core_filters` live. Measured: `trial.locationtrial_set.count()` = 6, `.using('trials').count()` = **43021**, the whole table, with no exception and no log line. Fixed with `db_manager(alias).get_queryset().using(alias)`.
- **Nothing tested that `install()` is ever called.** Deleting it from `TrialsConfig.ready` left all 922 tests green — the only wire between the module and Django, untested, in the very commit whose subject is making the off-switch test capable of failing.

One gap is left open deliberately and documented in the code: the trailing `.using(alias)` is load-bearing (a prefetched relation's `get_queryset` returns the cache and ignores `_db`) but has **no test**. Every in-process attempt to build the distinguishing case — prefetch on A, then `.using(B)` — failed to reproduce, so the attempt was deleted rather than kept as a test that cannot fail. It stands on a corpus measurement, and the comment says to reproduce that before touching it.

## What actually closes #360

Nothing here does, and nothing in this design would. In codex's order of preference:

1. **Ask the corpus owner to add and backfill the column.** The real fix — `?therapy_id=` cannot work correctly without the data, not merely without the column.
2. **Add it to our own copy** — `jsonb NOT NULL DEFAULT '[]'` plus the GIN index. List and detail recover immediately.
3. **A compatibility view** synthesising `[]::jsonb AS omop_intervention_concept_ids`, giving Django the row shape it expects with no manager patching.
4. Only then application code: defer the field, skip it in the reflective detail walker, and make `?therapy_id=` fail **explicitly** rather than 500 — and not silently return an empty list, which would be a semantic lie.

A `.defer()` alone reaches 200 on `GET /trials/` and leaves the detail page and `?therapy_id=` at 500, so it is item 4 done halfway. #360 stays open until each documented behaviour has an intentional outcome.

PR #363, opened 2026-08-11
Harden PHR auth: opt-in introspection, alg pinning, RFC 7662, CD trigger fix

LeoMo42:
Stacked on `feat/phr-federation` (#359) — addresses findings from a two-part self-review (fresh Claude subagent + `codex review`) of that PR.

## What this fixes

**H1 (High) — unauthenticated DoS / amplification via introspection.** In #359 any token with `alg ≠ RS256` and `iss=PHR_ISSUER` (gated only on the *unverified* issuer) fell through to a blocking `httpx.post` introspection call, uncached. `PHR_INTROSPECT_URL` was auto-derived from `PHR_BASE_URL`, so this was live in RS256 production, not just HS256 dev.
- Route only `RS256 → offline JWKS` and `HS256 → introspection`; every other alg (`none`, `RS384/PS256`, HS256-for-RS256 confusion) is rejected instead of routed to an outbound call.
- Introspection is now **opt-in**: `PHR_INTROSPECT_URL` is no longer derived from `PHR_BASE_URL`. A public RS256 deployment leaves it empty → no unauthenticated introspection surface. `docker-compose` sets it explicitly so local HS256 federation is unchanged.

**L2 / M1 — introspection request hardening.** RFC 7662 §2.1: form-encoded body (was JSON) + optional client auth via `PHR_INTROSPECT_AUTH` (`id:secret` → HTTP Basic; a value without `:` is ignored with a warning rather than sent as an empty password).

**P1 — CD deploy never triggered.** `deploy-render.yml` listened for `workflow_run.workflows: [django]`, but the CI workflow is `name: backend`; `workflow_run` matches the display name, so the Render deploy never fired. Now `[backend]`.

## Not changed (deliberate)
- **Default `PARTNER_AUTH_PROVIDERS` (review M3)** left as both providers — flipping to Firebase-only risks disabling PHR on the Render service whose env isn't visible in-repo; the opt-in introspection fix already removes the DoS surface that made the broad default dangerous.
- **`aud` validation (M2)** / **`TRIALS_DATABASE_INIT_FROM_BACKUP` drop-schema (L4)** — noted as follow-ups; L4 is pre-existing and not touched by #359.

## Tests
New `tests/test_phr_provider.py` (11 adversarial tests): alg routing (`none`/`RS384` rejected without any network call), opt-in no-network guard, RS256 valid/refresh/wrong-issuer, HS256 introspection active/inactive/refresh, form-encoding + Basic-auth assertion, response-subject authority. The no-network guard records calls and asserts **after** `verify()` returns (a raise inside `_verify_introspect`'s `except Exception` would be swallowed), so it genuinely fails if the routing/opt-in hardening is reverted. `tests/test_auth.py` + new file: 18 passed.


LeoMo42:
Conflicts resolved — `mergeStateStatus: CLEAN`, CI green. Resolution merge is `31a230c`.

## Why it conflicted, and what is left of this PR

The base moved when #364 landed, and #364 did overlapping work. Going through it item by item, three of this PR's four contributions are now superseded:

| this PR | state |
|---|---|
| alg pinning | base has it, with a named `_INTROSPECTABLE_ALGS` and an `isinstance(alg, str)` guard |
| opt-in introspection | base has `PHR_ALLOW_INTROSPECTION`, which fails closed on an unset `ENVIRONMENT`, plus pre-network access/expiry checks and a call budget |
| CD trigger fix (`workflows: [backend]`) | already in the base, with the commit pinning and fork-PR guard on top |
| **RFC 7662 §2.1 form-encoding + client auth** | **not present anywhere** |

So the merge takes the base's version at every conflict and re-applies the fourth item by hand.

One correction to my own reasoning while resolving: I first wrote that this branch's `elif alg == "HS256"` "would raise TypeError" on a list-valued `alg`. It would not — `==` never raises. The TypeError risk is in the base's `in frozenset` form, which is exactly why the base needs the `isinstance` guard it has. The two are equally type-safe; the base wins on being the base, on the named constant, and on admitting HS384/HS512. All three were untested in either direction, so widening or narrowing that set was free — there is now a test for it (narrowing back to HS256 fails two cases).

## The surviving item is worth keeping

The base posts `json={"token": token}`. RFC 7662 §2.1 specifies a form-encoded body, so a spec-compliant portal parses this as a form, finds no `token` parameter, and answers "not active" — for a token that is perfectly good. A rejection that looks like a verdict.

Now form-encoded with `token_type_hint`, plus optional HTTP Basic from `PHR_INTROSPECT_AUTH`. `partition` on the first colon keeps a secret containing a colon intact. A value missing *either* half — `"id"`, `"id:"`, `":secret"`, `":"` — is refused with a warning rather than sent as a half-blank credential, which would read as authenticated in the portal's log and is not. (The original guard tested for "a colon anywhere", which would have let `"id:"` straight through the check written to stop it.)

Plus an explicit empty-URL check, placed with the other no-network checks and **ahead of the call budget**. Behind the budget, an unconfigured deployment spends its whole allowance on calls that were never going to happen, so the first properly configured call gets refused by an exhausted budget. Its test now asserts that, not just the absence of calls.

## `tests/test_phr_provider.py` is dropped rather than repaired

All eleven of its tests are covered by `tests/test_phr_auth.py`, which the base added and which is considerably more thorough on the same ground — and three of them failed after the merge because they predate the `exp` requirement and the introspection gate. Two test files for one provider, with different fixtures, one written against superseded behaviour, is a trap rather than coverage.

Its only uncovered assertions — the request shape — move into `TestIntrospectionRequestShape` beside the introspection fixtures. Worth noting: that fixture recorded `kwargs['json']`, so after this change it would have recorded `None` for every call and silently stopped asserting anything about the wire format. It now records the whole call.

## Docs that would otherwise have gone stale

`docs/setup.md` still said `PHR_BASE_URL` derives `PHR_INTROSPECT_URL` and listed the latter as `_(derived)_`. Following that table, an operator sets `PHR_ALLOW_INTROSPECTION=true`, watches every HS256 sign-in 401, and gets a **debug-level** line as the only symptom. `.env.example` mentioned neither variable. Three code comments still gave "both URLs derive" as the *reason* the gate exists — including the docstring of the test class whose whole job is defending that gate's default.

## Verification

Both halves of the review gate, run to convergence. Everything is mutation-checked rather than merely passing:

- revert the request block to `json=` → the wire-format tests fail
- restore "a colon anywhere" → three credential cases fail
- move the URL check behind the budget → its test fails
- narrow the alg set to HS256 → two cases fail

Suite **880 passed / 6 skipped**, against **868** on `origin/feat/phr-federation` alone; the +12 is exactly the tests added here. `makemigrations --check` clean.

## One thing to decide

This PR targets `feat/phr-federation`, which reaches `main` via #359. Since #414 was retargeted to `dev` and now carries the stack, **merging this will not reach `dev` on its own** — I checked, and `31a230c` is not an ancestor of `feat/mf-bridge-trialmatches`, while `feat/phr-federation`'s current tip is. After this lands, the chain needs merging up into #414's branch again, or `dev` ships the introspection path with the JSON body.

PR #365, opened 2026-08-12
Keep the schema-drift tolerance off the primary database, and test the mechanism

LeoMo42:
Targets `fix/legacy-trials-schema` (#361), not `main` — merge this first and #361 carries it.

Review of #361 found that the tolerance could degrade the one database it was never meant to touch, and that the mechanism it exists for had no test at all. Three further rounds on the fix itself found more, including a regression the fix introduced.

## The corpus, and only the corpus

`install()` wrapped every manager in the app and `defer_missing_columns` keyed off `queryset.db`. The router sends trials models to `default` whenever `TRIALS_DATABASE_URL` is unset, so with the flag on in single-database mode the deferral applied to the **primary** database. Reproduced locally, where the app's own database happened to be two columns behind:

```
trials_trial on 'default' is missing 2 column(s) the models declare:
lesion_size_mcl_min, lesion_size_mcl_max. Deferring them; see exact#360.
```

That is an un-applied migration, not an externally owned corpus. With the `details()` skip from #361's second commit the degradation reached clients too: `POST /trials/<id>/match/` answers **200 with a short eligibility table** where it used to raise, so an operator who forgot `migrate` sees green.

Closing it took three attempts, and the first two are worth recording because both were the same mistake — trusting configuration text:

1. Gate on the alias name. Broken by pointing `TRIALS_DATABASE_URL` at the same host, port and name as `default` — a plausible way to exercise the split-database path against one database. Measured: the tolerance deferred those same two columns, now "legitimately", under the name `trials`.
2. Gate on host/port/name. Wrong in **seven** measured configurations — a URL stating no port against an explicit `5432`; `localhost` against `127.0.0.1`, though `localhost` is this project's own default `DATABASE_HOST`; a different case; a trailing root dot; a local socket directory under peer auth; a hostname against the primary's IP; and, in the other direction, two *different* Cloud SQL instances falsely matching, because `dj_database_url` leaves `HOST` empty and puts the socket path in `OPTIONS` — which made the feature silently inert exactly where a socket-form deployment needs it.

Host spellings are unbounded, so the guarantee no longer rests on them. A corpus is **defined** at the top of `db_compat.py` as a database with no `django_migrations` table, and that is now the check: if the target has one, this is a database the project migrates and nothing is deferred. It costs nothing — the table list is already in hand for the missing-table branch — and it subsumes every misspelling at once. The config comparison stays as a cheap pre-check so `install()` can decline without touching a database, and its docstring says it is not the guarantee. A corpus carrying a stray `django_migrations` cannot use the tolerance; the refusal says so.

## A regression the fix introduced

`connection.introspection.table_names()` excludes views by default (`ti.type == "t"`, and Postgres types `relkind IN ('m','v')` as `"v"`). A legacy corpus published under Django's table names is very plausibly exposed as views over another schema — so the pre-check read a present relation as an absent table, cached that permanently, and 500'd every endpoint with the flag armed. The replaced `information_schema.columns` query included views, so this was a regression in the target scenario. Now `include_views=True`.

## Manager patching

A per-manager-class subclass assigned to `manager.__class__` on `model._meta.local_managers`, instead of a `types.MethodType` on the instance. `db_manager(alias)` returns `copy.copy(self)` with `_db` set, and an instance-level method stays bound to the original, so the copy's alias was silently ignored — `Trial.objects.db_manager('trials')` queried `default`. The same change makes the patch survive `apps.clear_cache()`, which discards the `Options.managers` cached_property and took any per-instance patch with it, failing back to 500 with no signal.

`Manager.using(alias)` needed the same care: it is a passthrough to `get_queryset().using(alias)`, so the deferral was decided against the router's alias and the query then moved elsewhere, carrying corpus deferrals onto `default`. It picks the alias first now.

The generated subclass is importable under the name it reports, so managers pickle again (they raised `PicklingError`) and a future `use_in_migrations` manager will not break `makemigrations` — on the pre-fix code `Disease.objects.deconstruct()` really did raise `ValueError`.

## Robustness

- Introspection failures are remembered for `INTROSPECTION_RETRY_INTERVAL` (60s), not forever. Neither extreme is right: uncached, an unreachable corpus costs a connection attempt and a full traceback **per queryset built anywhere in the app**; cached forever, one blip disables the tolerance until the worker restarts and every later query fails on the missing column.
- Alias resolution moved inside the `try` — `ConnectionDoesNotExist` was raised out of every queryset.
- `except Exception` narrowed to database errors. It had already absorbed `TypeError` from this module's own test stubs twice, turning two tests inert, and a Django rename of the introspection API would have made the tolerance permanently inert on a retry loop while endpoints 500 anyway.
- The alias name is imported from `exact.db_router` rather than restated, so the two cannot drift and leave this covering an alias nothing routes to.
- Successes are cached for the life of the process — documented, with "restart workers after the corpus is migrated", since bounding it would mean re-introspecting on a timer for a schema that almost never changes.

## Documentation that was wrong in both directions

The docstring justified patching every manager with `trials_therapy.omop_concept_id` being missing too. That column **does not exist on this branch**: `omop_concept_id` appears nowhere in `trials/` except that sentence, migrations stop at `0008_trial_omop_intervention_concept_ids`, and ADR 0001 records the OMOP therapy columns as unmerged. The reasoning was carried over from the `dev` model set.

The limits list now names what deferral cannot reach: `filter()`/`order_by()` — including the live `?therapy_id=`, which reaches `TrialQuerySet.by_therapy_id` and still fails against a corpus lacking the column — `values()`, `select_related()` (the joined model's columns come from its own field list, and the app uses it on trials models), and `Model._base_manager`, hence forward FK dereference.

It had also gone wrong the other way: the claim that related managers are *not* covered was false, and my correction overstated it. Reverse and many-to-many managers do inherit the deferral, confirmed directly — but only because `install()` runs before any descriptor caches its manager class. That is ordering, not a guarantee, and it is stated as such. Forward FK is not covered either way.

## Tests

**+~470 lines. 920 passed / 6 skipped. Every fix reverted one at a time; each fails.**

Thirteen tests in this file were incapable of failing, measured rather than assumed. Listing them all because they are one mistake in thirteen costumes, and because four were caught only after this branch was pushed — by a review round that read the commit message's own claims back against the code:

| what made it vacuous | how it was caught |
|---|---|
| `defer_missing_columns` reduced to `return queryset` | whole suite stayed green — the mechanism was untested |
| `connections` stubbed with a `property` object | its `TypeError` was swallowed into the asserted `()` |
| fixture named the corpus alias `'corpus'` | the router keys off `'trials'`, so every deferral under it was inert |
| stub raised inside `get_table_description` | absorbed by the module's own `except Exception` |
| `install()` off-switch test configured no corpus | it returned on the alias gate whatever the flag said |
| `db_manager('trials')` asserted | the router would have chosen `trials` anyway |
| `.count()` on a drifted table | selects no columns, so it passes against a table missing every one |
| `makemigrations --check` asserted | no trials manager sets `use_in_migrations`, so it never deconstructs them |
| off-switch test for `defer_missing_columns` | no corpus configured, so the alias gate returned early however the flag read — **and the first two fixes for it did not work either**, since with the flag off `missing_columns` returns before reading anything they patched, and with it forced on the database matched the models |
| `_base_manager` / `deconstruct` / `db_manager` tests | all three passed with `install()` reduced to a no-op: they assert properties that hold of unpatched managers too |
| `assert ABSENT not in out` in the `details()` test | that field is never rendered by `details()`, so the assertion could not fail |

The off-switch one is worth dwelling on: three separate rounds asserted it was fixed, and it was still vacuous each time. What finally works is configuring a corpus *and* stubbing drift into it, so the flag is the only thing left saying no.

There is now a class that drops the column inside the test transaction and runs real introspection and a real deferral against it: an undeferred fetch raises, a deferred one returns the row, the column leaves the `SELECT`, and reading, filtering and `values()` on it still raise. Plus guards on the migrations backstop, on both refusals, on the seven host/port/name configurations, on `db_manager('default')`, on pickling and `deconstruct()`, on managers built *before* `install()` being replaced, and on `_base_manager` deliberately not being covered.

`docs/setup.md`'s env table and `.env.example` gain the flag — and `TRIALS_DB_MIGRATE`, which was also missing — including that the two accept different value spellings.

## Still open, deliberately

- `_base_manager` is not wrapped, so following a forward FK into a drifted table still 500s. Fixing it means `base_manager_name` on 56 models or patching `Options.base_manager` — too much for a stopgap.
- `GET /trials/?therapy_id=…` still fails against the `main` corpus: deferral cannot help a `filter()`. #360 is the fix.
- Nothing tells an operator the flag is active beyond one warning per model per process.
- A corpus separated from `default` by a role `search_path` rather than by host is refused by the config pre-check, even though the relations it exposes are not the ones `migrate` maintains. Refusing is the safe direction; recorded in the docstring.


LeoMo42:
Heads-up from the other side of #360: there are now two branches that fix the same failure by making the models a **subset** of the catalog instead of tolerating drift at runtime — `harvard/catalog-schema-fix` (base `main`, tagged `harvard-2026-08-14`) and `align/cb-catalog-schema` (base `2omop`). They remove `Trial.omop_intervention_concept_ids` and its GIN index, `by_therapy_id`, `StudyPreferences.therapy_id`, `RawDataItem.old_raw_data`/`.extracted_data`, the `Marker`/`MarkerCategory`/`MarkerCategoryConnection` trio (replaced by upstream CB's `MolecularMarker` + `CytogenicMarker`) and three models nothing queries.

Measured against a 2026-08-14 CB production dump and against that dump run through the Harvard pruning script: `MISSING TABLES (0)`, `MISSING COLUMNS (0)`, 531 columns compared with no type or nullability mismatches, every endpoint 200 with zero DDL on the catalog.

**This collides with the tests here, in either merge order:**

- `tests/test_db_compat.py:21-24` — `ABSENT = 'omop_intervention_concept_ids'`, and `TestTheDecision` asserts that field is still in `Trial._meta.local_fields` ("fixture assumes this field exists on Trial"). Once the alignment lands, that assertion fails before the mechanism is exercised at all. The real-DDL and detail tests create/drop/read the same column.
- `trials/db_compat.py:33-36` — the docstring names `?therapy_id=` / `by_therapy_id` as the live uncovered case. That path is being deleted (see #379 for why: EXACT-only column CB never populated, built for a demo that actually wanted a `therapy_concept_id`).

Suggested fix on this branch: use a synthetic column, or a field the subset rule will not touch, rather than a real EXACT-only one — a fixture that depends on a column we are actively removing will keep breaking.

**Not asking anyone to drop this work.** The two mechanisms cover different failure classes and both are worth having:

| failure | subset alignment | this tolerance | fresh transfer |
|---|---|---|---|
| EXACT declares a column CB will never have | removes the cause | hides it on read | n/a |
| catalog lags a column during a CB rollout | no help | **this is the case for it** | helps |
| catalog lags a whole table + data | no help | out of scope by design | **only fix** |

Worth stating the split explicitly in the PR body so the flag does not become the default: subset + the CI check proposed in #373 is the norm, this is an opt-in emergency read tolerance for missing *columns*.

LeoMo42:
## The one question that decides both PRs, measured

#365's `django_migrations` backstop is the first guard here that does not depend on configuration text, and a fresh-context review could not defeat it — every miss it found fails *closed*. That is good engineering. It is also why I went and checked what our corpus actually looks like.

**Every trials corpus on this machine carries `django_migrations`:**

```
exact_mm             django_migrations present (41 rows)   3114 trials
exact_multi          present (51)                         18462
exact_omoptest       present (51)                         13067
exact_shadow_221     present (51)                         13067
exact_legacy_corpus  present (31)                             1
```

`exact_legacy_corpus` is the byte-copy of the actual drifted corpus this work exists for.

End to end, with the flag ON and `TRIALS_DATABASE_URL` pointed at that copy:

```
флаг включён:      True
алиас корпуса:     trials
отложено колонок:  ()
```

and the log line:

> `'trials' has a django_migrations table, so it is a database this project migrates rather than an externally owned corpus. Deferring nothing… (A corpus that carries a stray django_migrations cannot use this tolerance.)`

So on the database this was written for, **#361 + #365 defer nothing.** The corpus is a dump of the upstream CancerBot Django database, and a full dump carries that table — which is the normal way to produce it.

## What does work, measured on the same database

```
really missing:                             ['omop_intervention_concept_ids']
Trial.objects.all()                          -> ProgrammingError
Trial.objects.defer('omop_intervention_…')   -> OK
```

One column, read nowhere, closed by one `.defer()` on the trials queryset. No flag, no introspection, no manager subclassing, no guards to keep ahead of configuration text.

**Correction to something I said earlier in this review thread:** I proposed hoisting the unconditional `getattr` in `trial_attributes.py`. That does **not** work — `omop_intervention_concept_ids` is a plain field in the `SELECT` of every trial fetch, so the query dies before any attribute is read. The conclusion survived; the mechanism I named did not. Credit to the review that measured it rather than taking my word.

## Recommendation

Close both, and land the `.defer()` against #360. Keep `trials/db_compat.py` on the shelf for the day a corpus drifts by more than one unread column — the two-checks-plus-backstop design is the right shape *if* runtime schema tolerance is ever genuinely needed.

Before closing, the two P1s the review of #365 found are fixed regardless (incoming PR) — the branch should not sit with a known wrong-results bug even if it is then shelved:

- `Deferring.using()` dropped the relation filter on related managers. Measured on the corpus: `trial.locationtrial_set.count()` = 6, `.using('trials').count()` = **43021**, the whole table, with no exception and no log line. A regression #365 introduced by moving from per-instance patching to a subclass, which put `Deferring` in the middle of every related manager's MRO.
- Deleting `install(self)` from `TrialsConfig.ready` left all 922 tests green — the only wire between the module and Django, untested.

LeoMo42:
Closing. The measurement that settles it is in the comment above: **every trials corpus carries `django_migrations`, including a byte-copy of the real drifted one**, so with the flag on and the alias correctly configured this defers zero columns and logs a refusal. #361 + #365 are inert against the database they were written for.

I cross-checked the decision with codex rather than only my own reading. It agrees on closing, and corrected me on three things worth recording here:

1. **"The column is read nowhere as an attribute" was too strong** — the generic detail-field walker reads it indirectly. That is why the detail page keeps 500ing whatever the SELECT list says.
2. **"Keep `db_compat.py` on the shelf" is questionable.** Its limits are *structural*, not unfinished polish: if drift ever touches filters, relations, reflective readers or related models it fails unpredictably. Worth preserving the design discussion — not this implementation as the assumed future answer.
3. **#360 must stay open.** A partial fix that closes the issue hides the remaining failures and makes the workaround permanent by accident.

## The two P1s are fixed regardless — `b20c5be`, branch `fix/365-related-manager-using`

Pushed rather than PR'd, since this branch is closing. The work should not be lost and the branch should not sit with a known wrong-results bug.

- **`Deferring.using()` dropped the relation filter.** Moving from per-instance patching to a subclass put `Deferring` in the MIDDLE of every related manager's MRO, so `super(Deferring, self).get_queryset()` skipped `RelatedManager.get_queryset` — where `core_filters` live. Measured: `trial.locationtrial_set.count()` = 6, `.using('trials').count()` = **43021**, the whole table, with no exception and no log line. Fixed with `db_manager(alias).get_queryset().using(alias)`.
- **Nothing tested that `install()` is ever called.** Deleting it from `TrialsConfig.ready` left all 922 tests green — the only wire between the module and Django, untested, in the very commit whose subject is making the off-switch test capable of failing.

One gap is left open deliberately and documented in the code: the trailing `.using(alias)` is load-bearing (a prefetched relation's `get_queryset` returns the cache and ignores `_db`) but has **no test**. Every in-process attempt to build the distinguishing case — prefetch on A, then `.using(B)` — failed to reproduce, so the attempt was deleted rather than kept as a test that cannot fail. It stands on a corpus measurement, and the comment says to reproduce that before touching it.

## What actually closes #360

Nothing here does, and nothing in this design would. In codex's order of preference:

1. **Ask the corpus owner to add and backfill the column.** The real fix — `?therapy_id=` cannot work correctly without the data, not merely without the column.
2. **Add it to our own copy** — `jsonb NOT NULL DEFAULT '[]'` plus the GIN index. List and detail recover immediately.
3. **A compatibility view** synthesising `[]::jsonb AS omop_intervention_concept_ids`, giving Django the row shape it expects with no manager patching.
4. Only then application code: defer the field, skip it in the reflective detail walker, and make `?therapy_id=` fail **explicitly** rather than 500 — and not silently return an empty list, which would be a semantic lie.

A `.defer()` alone reaches 200 on `GET /trials/` and leaves the detail page and `?therapy_id=` at 500, so it is item 4 done halfway. #360 stays open until each documented behaviour has an intentional outcome.

PR #395, opened 2026-08-30
fix(matching): decide the therapy count from the same bag the matcher reads (follow-up to #394)

LeoMo42:
Follow-up to #394: the count has to ask the therapy **bag**, not the `first_line_therapy` column.

## What #394 missed

`_match_therapy_lines_attr` answers `first_line_therapy` from `get_user_therapies()`, which folds in `second_line_therapy`, `later_therapy`, `later_therapies` and the supportive rows. `is_attr_blank` looks only at the one column. #394 gated its scoped fragment on that column being non-blank, so the two paths still disagreed whenever the value that drives the verdict sits on another line:

- **unexpandable value elsewhere** — the count kept the old all-six-columns fragment and made a regimen-only trial Potential while the matcher called it eligible;
- **expandable value elsewhere** — the count read `first_line_therapy` as unanswered and demoted a trial the matcher had already decided, `matchScore` dropping to match.

Both measured with the comparator before fixing, both now pinned:

```
blank + expandable(second)      therapies_excluded            sql=potential  matcher=eligible  DIVERGE
blank + expandable(supportive)  therapy_components_required   sql=potential  matcher=eligible  DIVERGE
```

## The change

The branch keys on the bag being non-empty, and sets `is_filled_by_user` from whether it expands: expands → answered (the matcher decides every leg from it), does not expand → the two scoped fragments as before. An empty bag is still left to `is_attr_blank`, the state both paths already agree on.

## Verified

- 8 new cases across the scalar lines, `later_therapies`, the supportive rows, the empty-bag control, and the not-eligible leg (where the hard filter is reached through the post-loop fire rather than the therapy-line dispatch). Non-vacuous: 6 of them fail against #394 as merged.
- Full suite: **1 373 passed, 5 skipped, 1 xfailed**.

## Same fix, other repo

CancerBot runs its own copy of this mapper, so the identical correction is in cancerbot-org/cancerbot#5052. Keeping the two in step is the point; letting only one of them learn this would be how the next divergence starts.

## Review status

Fresh-context review done (it is what found the expandable half) and its findings applied. **Codex's half is owed on the final state** — quota exhausted until ~18:16; result will be posted here before this should be merged.

### DIFF
--- a/matcher_app/exact_matching/user_to_trial_attrs_mapper.py
+++ b/matcher_app/exact_matching/user_to_trial_attrs_mapper.py
@@ -310,7 +310,15 @@ class UserToTrialAttrsMapper:
                 # `eligible_for_therapy_related_things_from_lines` still runs on the
                 # raw values, so a value that contradicts a required regimen keeps
                 # excluding the trial.
-                if is_filled_by_user and user_attr in THERAPY_LINES_ATTRS_UNDERSCORED \
+                # Not gated on `is_filled_by_user`: the matcher answers
+                # `first_line_therapy` from `get_user_therapies()`, which folds in
+                # the other therapy lines and the supportive rows, so an
+                # unexpandable value in `second_line_therapy` leaves this field
+                # blank while still driving the verdict. Deciding from the same bag
+                # is what keeps the two paths together —
+                # `_therapies_do_not_expand` returns False for an empty bag, the
+                # case `is_attr_blank` already owns.
+                if user_attr in THERAPY_LINES_ATTRS_UNDERSCORED \
                         and not has_no_prior_therapy:
                     if therapies_do_not_expand is None:
                         therapies_do_not_expand = self._therapies_do_not_expand(service)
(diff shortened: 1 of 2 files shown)


LeoMo42:
Codex's half is in. No blocking flaw — it confirmed that for a non-empty bag the count now uses the same source of truth as the matcher, and that `continue` and the `has_no_prior_therapy` bypass are correct. Its asks were coverage, and one of them found something:

- **mixed bag** (one resolvable therapy plus one junk one) — must count as answered, since the matcher derives from the resolvable one and is definite. Added.
- **`has_no_prior_therapy` × each bag state** — the branch deliberately bypasses that route, so it has to agree on its own. **It does not for one of the six**: a patient who says "no prior therapy" while carrying a resolvable one is dropped by the coarse no-prior filter (which keeps only trials requiring nothing) and called *eligible* by the matcher (which evaluates the real overlap). Pre-existing — neither #394 nor this touches that route — 4 such patients in the production catalog. Filed as #396 and `xfail(strict)` here, so closing it flips this green.
- **empty-bag test was weak**: it asserted only that the paths agree, not that the special fragment stayed out of it. Now asserts the generic six-column fragment is what fired.
- **the match-score denominator**: the attr now sits in both count dicts, and `match_score` divides by `num_nonnulls(potential + filled)`. The two fragments are mutually exclusive by construction, so it contributes exactly one — now pinned by a trial gating on a component *and* a regimen column asserting `potential_attrs_count == 1`.

Full suite: 1 381 passed, 5 skipped, 2 xfailed. Both halves of the review are satisfied on the current state.

Issue #397, opened 2026-08-30
Audit and restrict GCS IAM on the DB bootstrap restore sources


LeoMo42:
## Summary

The container bootstrap restores databases from objects in a GCS bucket. The IAM bindings and object ACLs on that bucket have never been audited, and the restore path has no integrity control. A security review flagged this as the highest-priority item in the repo; the specifics are being shared out-of-band rather than here.

## Where

- `docker/init_trials_db.sh:14` and `docker/init_patients_db.sh:14` — the default restore sources.
- `docker/entrypoint.sh:14-15` — both scripts are invoked on every container start; `Dockerfile.gcp:56` makes that the Cloud Run CMD.
- `docker/Image.md:86-87` — the published quick-start enables both `*_INIT_FROM_BACKUP` flags.

## What to do

1. **Audit the bucket IAM.** Enumerate every principal with read and with write on the bucket and on the individual objects, including any legacy ACLs and any `allUsers` / `allAuthenticatedUsers` binding. Write down what each one is for.
2. **Restrict to least privilege.** Whatever does not need access should lose it. Any data path that is not intended to be world-readable belongs in a separate private bucket, not alongside content that is.
3. **Confirm the deployed flag state.** `gcloud run services describe` on the staging and production services — verify whether `TRIALS_DATABASE_INIT_FROM_BACKUP` / `PATIENT_DATABASE_INIT_FROM_BACKUP` are set anywhere. Nothing in this repo sets them, but the service env is managed out-of-band.
4. **Add integrity checking to the restore.** The scripts download an object over HTTPS and pipe it into `psql` after `DROP SCHEMA IF EXISTS public CASCADE` (`init_trials_db.sh:50-67`). There is no checksum, no signature, no object-generation pin and no size ceiling, and in the documented compose the restoring role is a database superuser. Pin the object generation or publish and verify a `sha256`, and restore as a non-superuser role.
5. **Drop the hardcoded default URL** so the restore source has to be configured explicitly rather than inherited.

## Context

Full write-up is in the local security report at `.gstack/security-reports/2026-08-30-comprehensive.md` (findings 1 and 6). That path is gitignored and not committed — @LeoMo42 will share it and the supporting detail directly.

[...]

PR #399, opened 2026-08-30
chore(deps): Django 5.2.17, python-dotenv 1.2.3, requests 2.33.1

LeoMo42:
Part of the security-audit follow-up tracked in #397.

`pip-audit` reported **29 advisories against Django 5.2.7** — ten security releases behind the head of the same LTS line — plus one each for `python-dotenv` and `requests`. All three are patch bumps within their current minor; no API surface changes.

| | before | after |
|---|---|---|
| Django | 5.2.7 | 5.2.17 |
| python-dotenv | 1.2.1 | 1.2.2 |
| requests | 2.32.5 | 2.33.0 |

## This is patch currency, not an incident

Reachability was checked per advisory class rather than reported as a raw count. **No reachable exploit was found for any specific Django CVE in this configuration:**

| Advisory class | Reachable? |
|---|---|
| `FilteredRelation` SQLi | No — unused |
| `_connector` dict-expansion SQLi | No — every `Q(**{...})` uses static keys from `USER_TO_TRIAL_ATTRS_MAPPING`, `_connector` never passed |
| ASGI header spoofing / `Content-Length` bypass | No — gunicorn WSGI, no `asgi.py` |
| Cache-middleware `Vary` bugs | No — `UpdateCacheMiddleware` / `cache_page` unused |
| `GEOSGeometry` WKT recursion, `GDALRaster`, `RasterField` | No — only `Point(float, float)`, no WKT/WKB or raster parsing from request data |
| Admin `list_editable` / inline add-permission | No — `list_editable` unused |
| FS-storage / file-cache race | No — LocMem or Redis, no file cache |

The bump is worth doing because an auditor tests patch currency, and because it picks up the fixes for free.

## Verification

- Clean venv on the bumped pins: **1381 passed, 5 skipped, 2 xfailed**
- `pip-audit` after the bump: one advisory left, `pytest` 8.4.2 (dev-only, fixed in 9.0.3) — left for a separate change since it touches the test toolchain
- Codex reviewed the upgrade risk against actual usage: Django 5.2.17's only incompatible GIS change rejects `str`/`dict` raster values in spatial lookups, which this code never does; `requests` 2.33.0 drops Python 3.9 (we run 3.12) and changes nothing used by the PROMOP clients; `python-dotenv` 1.2.2 changes only `set_key`/`unset_key`/CLI behaviour, and `exact/settings.py` uses `load_dotenv` alone.
- `matcher_app/pyproject.toml` allows `Django>=5.0`, so nothing else pins these.

## Follow-up

Consider adding `pip-audit` to `django.yml` so this gap cannot silently reopen.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,10 +1,10 @@
 # Core
-Django==5.2.7
+Django==5.2.17
 djangorestframework==3.16.1
 djangorestframework-camel-case==1.4.2
 drf-yasg==1.21.14
 django-cors-headers==4.9.0
-python-dotenv==1.2.1
+python-dotenv==1.2.3
 gunicorn==23.0.0
 whitenoise==6.11.0
 
@@ -33,7 +33,7 @@ firebase-admin==7.4.0
 django-countries==8.2.0
 glom==24.11.0
 inflection==0.5.1
-requests==2.32.5
+requests==2.33.1
 
 # Testing
 pytest==8.4.2


LeoMo42:
Updated after the second review pass: `requests` goes to **2.33.1** and `python-dotenv` to **1.2.3**, not 2.33.0/1.2.2.

`requests` 2.33.0 still carries the `_parse_content_type_header` bug where a `Content-Type` parameter with no `=` yields a bool, so `Response.encoding` becomes `True` and `iter_lines(decode_unicode=True)` raises `AttributeError`. That is a live path here — `vocab_mirror/promop_vocab_client.py:172` streams the NDJSON vocab snapshot exactly that way, and the error escapes outside the `except requests.RequestException` guard at :178, surfacing as an unhandled error rather than a `VocabSyncError`. Verified on 2.33.1:

```
>>> _parse_content_type_header('application/x-ndjson; charset')
('application/x-ndjson', {})
```

2.33.1 also tightens the header validators from `$` to `\\Z`, which matters because both promop clients build `Authorization: Bearer {token}` from a remote OAuth JSON body.

`python-dotenv` 1.2.3 over 1.2.2 for a similar reason: 1.2.2 fixes nothing we use, 1.2.3 fixes a UTF-8-BOM `.env` silently dropping the first variable.

Suite re-run on the updated pins: 1381 passed, 5 skipped, 2 xfailed.

PR #414, opened 2026-09-08
Expose TrialMatches through a Module Federation bridge

vtrv101:
Exposes TrialMatches as a Module Federation bridge so a non-React host (HealthTree ONE) can mount it.

## Bridge

`TrialMatchesBridge` takes the service URLs and a token-getter and resolves the signed-in patient itself — `/patient-info/me/` against PRomop, then `/normalize-ctomop-row/` — so the host passes no clinical identifiers. Only JSON-serialisable values and functions cross the boundary.

Everything this remote injects now goes into the host's `mf-remote` cascade layer. This sheet is scoped under `.exact-root` and ships no Tailwind layers, so it was not the source of the host-chrome regressions the other remotes caused, but the contract is the same for every remote.

Context: `ht-phr/docs/mf-bridge-proposal.md`.

## Note on the diff

This branch was cut from the in-flight `feat/phr-federation` / `fix/legacy-trials-schema` stack (#359, #361, #365), so against `main` it also shows those four commits. Only the last three are mine. Happy to retarget it onto whichever of those lands first.

https://claude.ai/code/session_015dVAjtcAhgRnzmzFpYgWww


LeoMo42:
## Two-part self-review

Ran the pre-merge gate on this branch: an independent fresh-context review plus `codex review`, both against `2edbe62..32ece56` (this PR's own three commits), with a lighter pass over the four inherited stack commits. Both halves landed on the same lifecycle/state issues; the P0 below I re-verified by hand against the installed package.

**Verdict: two blockers, one of them in the API this PR publishes.** Everything is small to fix. Details:

---

### P0 — the documented `render`/`destroy` contract leaks the React root

`frontend/src/federation/TrialMatchesBridge.tsx:8-10` (and the `9e02a8e` commit message) tell hosts to call:

```js
await provider().render({ dom, baseUrl, getToken, ... });
provider().destroy({ moduleName, dom });
```

`createBaseBridgeComponent` returns `() => { const rootMap = new Map(); ... }` (`@module-federation/bridge-react/dist/bridge-base-*.mjs:39`) — **every `provider()` call gets its own `rootMap`**. `destroy` does `rootMap.get(dom)` and no-ops when absent (lines 142, 159). Measured against the real package:

```
documented usage      -> roots created: 1  unmounted: 0
single-instance usage -> roots created: 2  unmounted: 1
```

Failure: HealthTree ONE navigates away from Find Trials, calls `destroy` exactly as documented, and nothing unmounts. The tree keeps running, including `TrialMatches`'s `window.addEventListener("popstate", …)` and its `QueryClient`; the next mount calls `createRoot()` on a container that already has a live root, so React 19 warns and you get two trees on one node. Repeated navigation compounds it.

Fix: either document `const p = provider(); p.render(…); p.destroy(…)`, or memoize the factory so `provider()` is idempotent. Worth checking whether `ht-phr/docs/mf-bridge-proposal.md` §6.2 specifies the double call — if so, it needs the same correction there.

### P0 (targeting) — merging into `main` drops the hardening already written for the inherited stack

This branch carries four commits from the `#359`/`#361`/`#365` stack, but none of the fixes since written against them:

| commit | what | in this branch |
|---|---|---|
| `caa9308` (#365) | never tolerate schema drift on the primary database | NO |
| `a60bbd5` (#364, already in `feat/phr-federation`) | gate the portal introspection fallback, bound its outbound calls | NO |
| `8027f3c` (#363, open) | alg pinning, opt-in introspection, RFC 7662 form-encoding | NO |
| `d7514d2` / `257b0ca` | don't deploy on a fork PR; confirm the deployed commit | NO |

So merging as-is ships the un-hardened PHR auth provider and `TRIALS_DB_TOLERATE_MISSING_COLUMNS` without the primary-database guard. CI is green on all five checks, which is what makes this invisible. Retargeting onto `fix/legacy-trials-schema-review` resolves it — the PR description already offers this.

### P1 — an explicit `patientInfo` prop does not win once a patient has been resolved

`TrialMatchesBridge.tsx:128`: `load.status === "ready" ? load.patientInfo : rest.patientInfo`. After a successful `/patient-info/me/` + normalise, `rest.patientInfo` is ignored for the life of the mount. Host resolves patient A, the user switches profile (or the host starts impersonating) and re-renders with an explicit `patientInfo` — the bridge keeps matching against A. In a PHI context that is a wrong-patient display, not a stale render.

Paired with it (`:85-88`, found independently by `codex review`): the `useState` initializer only runs on first mount, so when `shouldLoad` flips to false the state stays `loading`/`error` and the effect returns early — the bridge is stuck on the spinner or the error card. codex rated this P2; the pair is P1 because of the wrong-patient outcome.

Fix for both: derive from props each render (`rest.patientInfo ?? fetched`) and reset `load` when `shouldLoad` or its inputs change.

### P1 — the cascade-layer wrap changes rendering for the existing `./TrialMatches` export

Per the cascade-layers spec, *unlayered* normal declarations beat **every** layer, regardless of specificity. Before this commit the remote's rules were unlayered and appended last to `document.head`, so they won ties. Now that they are all inside `@layer mf-remote`, any unlayered host declaration wins — e.g. an unlayered Tailwind v3 preflight (`button,[type='button']{background-color:transparent;background-image:none}`, specificity 0,0,1) now outranks `.exact-root .exact-btn-view` (0,2,0).

Two consequences:

1. `cssLayer.ts:19-21` asserts that a host declaring no layer order is a no-op ("nothing regresses"). That isn't right — the no-layer host is exactly the case that regresses.
2. "`./TrialMatches` is untouched" holds for the file, not for the result. ht-phr consumes that export today and gets the new cascade position without opting in.

Worth weighing against what the layer buys here: every selector in `frontend/src/federation/exact.css` is already scoped under `.exact-root` (all 60 checked), and the Tailwind-utilities leak that erased the HealthTree ONE header was removed in `2302352`, before this PR's base. The commit message concedes the sheet "was not the offender." Keeping the layer as a forward contract for other remotes is a reasonable call — but it should be made deliberately, and the comment should not claim the no-layer host is unaffected.

---

### P2

- `:78-81, 114` — `getToken` in the memo/effect deps. No internal loop (verified: `setLoad` re-renders with the same props object, so identity is stable). But a host passing an inline arrow on each `render()` produces a new `apiClient` and refires the effect — two extra round trips, one a POST of the full patient row, nothing debouncing it. Hold it in a ref and drop it from both dep arrays.
- `:51-54, 93` — no request timeout on either client, while `src/dev/exactAuth.ts:78` and `ctomopClient.ts:65` both set `DEFAULT_TIMEOUT_MS`. A hung PRomop leaves "Loading your health profile…" up forever, with no retry and no escape.
- `:93` — PRomop's API base path is hardcoded `"/api"` while EXACT's is a prop. In-repo evidence is contradictory: the dev harness uses `/api/patient-info/`, and this same branch puts the PHR portal at `/api/v1/…` (`exact/settings.py`). A mismatch gives a 404 behind the generic error card.
- `:56-61, 93` — one `getToken` is sent as `Bearer` to two different origins with no way to scope it. Given `accounts/providers/phr.py:184` sets `options={"verify_aud": False}`, cross-service token reuse isn't hypothetical in this family.
- `:106` — bare `catch {}` collapses 401 / 404 / 5xx / network / normalise failure into one message with no logging. "Please refresh to try again" is wrong advice for most of those.
- `:96-98` — an empty PRomop profile resolves to `patientInfo: null` and falls through to `TrialMatches.tsx:172-176`, which renders the developer message "Pass a `patientInfo` payload or `personId`…" — with literal `<code>` tags — to every user who hasn't finished onboarding.
- `cssLayer.ts:28` — `@import` is not in `HOISTED`, so it gets wrapped into the layer, where it is invalid and silently dropped. Meanwhile `@charset`, which browsers ignore inside `<style>` anyway, is in the list. Latent today (`exact.css` has no imports); becomes real the day someone adds a font or a shared sheet.
- `cssLayer.ts:80-84` — a stray `}` ejects the rest of the sheet *out* of the layer: `.a{color:red}}.b{color:blue}` yields `.b` outside `mf-remote`, i.e. at higher priority than everything inside. The `depth = 0` clamp keeps the parser going instead of failing loudly. A three-line invariant (`rules.join("") === css && depth === 0`, else return `css` unchanged) turns any future parser bug into a no-op.
- `cssLayer.ts:34-37` — hoisting inspects only the top level, so `@keyframes`/`@property` nested in `@media`/`@supports` are never hoisted; the latter is precisely the "not honoured in every browser" case the header comment says it guards. Also only `-webkit-` is matched, not `-moz-`/`-o-`.
- The host contract `@layer properties, theme, base, components, mf-remote, utilities;` is a new integration requirement for every host and lives only in a source comment — `frontend/README.md` documents the exposed modules and the mount flow but mentions neither it nor `./TrialMatchesBridge`.
- `TrialMatchesBridgeProps` is declared in `TrialMatchesBridge.tsx`, not in the exposed `./types`, so a TypeScript host cannot type its `render({…})` call against the contract.
- `TrialMatches.tsx:117-124` pushes `{ exactTrialDetail: id }` into `history.state` and listens for `popstate`. Inside a SvelteKit host that is now the only consumer path, and SvelteKit's client router keeps its own index in `history.state`. Untested, plausibly a hard navigation on back — worth exercising in the host.

### Nits

`apiBasePath` is concatenated with no normalisation (`"api"` → `https://hostapi`); `personId` stays in the prop surface but yields 403 by default (`EXACT_ALLOW_PERSON_ID_LOOKUP` off outside DEBUG); the loading state has no `role="status"`/`aria-busy` while the error state correctly has `role="alert"`; the injected `<style>` is never removed and `injected` never resets, which mount/unmount cycles now make routine; `:102` re-implements `api.ts:147`'s `normalizeCtomopRow`.

---

### Inherited stack — not this PR

Flagging only what's high-severity, and only because it would land with a merge into `main`:

- **P1** `accounts/providers/phr.py:198-203` — the introspection path has no rate limit, unlike the JWKS cache directly above it. Reaching this provider needs only an *unverified* `iss` claim (`can_handle`, line 141), and any `alg` other than RS256 falls through to a blocking `httpx.post(..., timeout=5)`. N forged `{"alg":"HS256","iss":PHR_ISSUER}` tokens produce N blocking outbound POSTs from a sync worker pool. `_JWKSCache` got `PHR_JWKS_MIN_REFRESH_INTERVAL` for exactly this reason; its introspection twin got nothing. Likely covered by `a60bbd5`/#363 — worth confirming that hardening addressed the DoS shape and not only the trust question.
- **P2** `phr.py:198` — the bearer token is POSTed to `PHR_INTROSPECT_URL` with no client authentication and no allowlist on the URL, so a misconfigured `PHR_BASE_URL` becomes a token exfiltration channel.
- **P2** `phr.py:182` — `options={"verify_aud": False}` with no audience check: any RS256 token the portal issued for any relying party authenticates here.
- **P2** `trials/db_compat.py:118-121` — `defer_missing_columns` reads `queryset.db` before `.using()` is applied, so the missing-column set is computed for the router-default alias rather than the requested one.
- **nit** `exact/health.py` inherits `AnonRateThrottle` (60/min, keyed on `REMOTE_ADDR`), so behind a proxy without `X-Forwarded-For` handling `/healthz` can 429 and the load balancer pulls a healthy instance. Consider `throttle_classes = []`.

---

### Checked and clean

`react-dom/client` is auto-registered as a singleton by `@module-federation/vite` (confirmed in the built `_virtual_mf-localSharedImportMap-*.js`), so its absence from the `shared` map is fine. The 19.2.6 pin is consistent across `package.json`, the lockfile and `react-dom`'s peer range, with no `overrides` and no second copy. bridge-react (~48 kB) does not leak into the `./TrialMatches` chunk, so React hosts pay nothing for the new export. No setState-after-unmount and no double-fetch race — the `cancelled` flag is correct. `cssLayer` held up against 17 adversarial inputs (apostrophes and braces in comments, unterminated comments, escaped selectors, custom properties containing braces, unquoted `url(data:…{…})`, layer statements) and round-trips the real built 8.4 kB sheet byte-for-byte. The standalone SPA is unaffected — it loads no global stylesheet. `vitest run` 34/34, `tsc -b` and the remote build both clean.

### Not verified

`ht-phr/docs/mf-bridge-proposal.md` §6.2 (outside this repo); whether ht-phr / HealthTree ONE hold the provider instance across render+destroy (determines whether the P0 is live or latent — the published doc is wrong either way); PRomop's real API base path and whether `/patient-info/me/` exists there; whether any host declares the `mf-remote` layer order (determines whether the cascade regression is live or latent); any runtime behaviour in a real host.

Fixes for the parts that are unambiguous defects are going up as a follow-up PR based on this branch.

LeoMo42:
Follow-up with the fixes is up as #415, based on this branch (`feat/mf-bridge-trialmatches`) rather than `main`, so it merges into this PR instead of racing it.

It covers the two P0s that are code (the `provider()` memoisation and the wrong-patient/stuck-state pair), the CSS-layer parser bail-outs, and the P2s. Two things surfaced while fixing that were not in the original list and are worth knowing about:

- The `console.warn` I proposed for the error path leaks the bearer token and the whole patient row if it is passed the `AxiosError` — `config.headers.Authorization` and, on the normalise call, `config.data`. #415 logs scalars only.
- `cssLayer` had a second silent-corruption case beyond the stray `}`: an unterminated comment swallows the layer's closing brace, so `.a{color:red}/*` becomes `@layer mf-remote{.a{color:red}/*}` and the block never closes.

Not addressed there, because they are not this PR's code: the base-branch question (#415 does not change it — this PR still needs retargeting off `main` before it lands, or it ships the stack without `caa9308`/`a60bbd5`/`8027f3c`) and the inherited-stack findings.

LeoMo42:
@vtrv101 — a question about where this should land, not about the code. You flagged the base yourself in the description ("happy to retarget it onto whichever of those lands first"), and while looking at it I found something that might make the choice easier.

**`dev` already solves everything the four inherited commits solve on the serving side, and solves it differently:**

| what `9b2c85b` adds here | `dev` today |
|---|---|
| Node stage in `Dockerfile` building the remote into the image | `Dockerfile.gcp` — `node:20-slim`, `npm ci`, `COPY --from=frontend … dist/remote` |
| whitenoise serving `/remoteEntry.js` with permissive CORS | `whitenoise==6.11.0` + `exact.middleware.CorsWhiteNoiseMiddleware`, `WHITENOISE_ROOT` |
| generated `frontend/package-lock.json` | present |
| pins react/react-dom to exactly 19.2.6 (part of `9e02a8e`) | `302ade5 fix(frontend): pin react + react-dom to exactly 19.2.6` |

So on `dev` the bridge needs none of the stack. I checked rather than assumed: cherry-picking just the six bridge commits (your three plus the three from #415) onto `origin/dev` applies **with no conflicts**, and on the result `tsc -b` is clean, 83/83 tests pass and `vite build --config vite.remote.config.ts` succeeds. The net diff against `dev` is purely additive:

```
+ @module-federation/bridge-react            (one new dependency)
+ TrialMatchesBridge.tsx, bridgeState.ts, cssLayer.ts + their tests
~ injectStyles.ts (1 line), vite.remote.config.ts (2 lines), types.ts, README.md
```

Nothing in `accounts/providers/phr.py`, `trials/db_compat.py`, the Dockerfile or the Render workflow. The react pin does not even appear, because `dev` already has it.

For contrast, merging this branch whole into `dev` conflicts on `docs/adr/0001-cross-vocabulary-mapping.md` (from `172a482`, one of the four commits `main` has and `dev` does not) and carries the PHR auth provider and `db_compat` along with it.

**Topology, since it is not a simple base flip:** `main` and `dev` diverged at `7631671` — `dev` is 104 ahead, `main` 4 ahead (two CI, two docs). This stack is cut from `7a367b3`, which is one of those four and is not in `dev`, so any move to `dev` is a real transplant rather than a retarget.

**The question, then:** where do you want the bridge to go?

- If `dev` is where this is wanted, the cheapest path looks like a fresh branch off `dev` with the six bridge commits and a new PR into `dev`, leaving this PR and the `#359`/`#361`/`#365` stack on their own `main` flow untouched.
- If it should stay on the `main` line, then nothing needs retargeting *provided the stack merges first* — the only real hazard is this landing in `main` ahead of `caa9308` (#365), `a60bbd5` (#364) and `8027f3c` (#363), which would put the un-hardened auth provider and an unguarded `TRIALS_DB_TOLERATE_MISSING_COLUMNS` into `main`.

Happy to prepare whichever you prefer — I have the cherry-pick verified locally already. Not moving anything until you say.

LeoMo42:
@vtrv101 

the default branch on the repo is `dev` -- Adam asked to freeze `main` branch for a while.

going to change the target branch now

LeoMo42:
Retargeted to `dev` and brought the branch up to date. CI green (build ×2, test 1/2/3), `mergeStateStatus: CLEAN`.

## What was merged in

- `cfac8ca` — `origin/fix/legacy-trials-schema-review` (#365 head), which docks the stack **with** its hardening: `caa9308` (never tolerate schema drift on the primary database), `a60bbd5` (gate the portal introspection fallback, bound its outbound calls), `d7514d2` / `257b0ca` (CI). Clean, no conflicts.
- `64dba2d` — `origin/dev`. One conflict, `docs/adr/0001-cross-vocabulary-mapping.md`, resolved to `dev`'s version.

`8027f3c` (#363) is **not** included — that PR is still open and lands on its own. Its substantive items turn out to be either already in HEAD via `a60bbd5` (alg allowlist, call budget) or superseded (`PHR_ALLOW_INTROSPECTION` is a stronger gate than not auto-deriving the URL). One thing is genuinely lost with it — see the follow-ups below.

## The ADR conflict was a false one

Same content, two SHAs. `main`'s `172a482` and `dev`'s earlier `50f5edb` produce the **identical blob** `1a0c136`; `dev` then has `fc65f9b` ("mark minimal-EXACT design as complete") on top → `f83577a`. So `dev`'s file is a strict successor of `main`'s and taking it loses nothing. HEAD carries `f83577a`.

Checked that nothing else was quietly changed under cover of the conflict: `git merge-tree` reproduces `cfac8ca`'s tree byte-for-byte, and the `dev` merge's auto-merge tree differs from the committed tree in exactly that one path. `git diff --diff-filter=D` against both parents is empty, and `git log <parent> ^HEAD` is empty for both, so every commit from both sides is an ancestor.

## One real integration bug, fixed in `6d9b013`

Both sides are green alone and the combination is not — the full suite came back **10 failed, 21 errors**, all in `tests/test_db_compat.py`, all `AttributeError: 'Settings' object has no attribute 'DATABASES'`, and none in code either side changed.

Measured:

| | |
|---|---|
| stack branch alone, full suite | 920 passed |
| `tests/test_db_compat.py` alone, on the merge | 52 passed |
| `tests/management/test_migrate_trials_db.py` → `test_db_compat.py` | **10 failed, 21 errors** |

The cause is on the `dev` side: `@patch('django.conf.settings.DATABASES', …, create=True)`. `create=True` makes `mock` skip the branch CPython labels *"needed for proxy objects like django settings"*, so `__exit__` takes `delattr` — which `LazySettings` forwards to the real `Settings` — instead of restoring. `DATABASES` and `TRIALS_DB_MIGRATE` are then gone for the rest of the session. It was invisible on `dev` because nothing after that file reads either setting; `test_db_compat.py`'s fixtures do.

Both settings exist in `exact/settings.py`, so `create=True` was never needed. Removed from all seven decorators (none left anywhere in the repo), with a note in the file so it doesn't come back. All four tests still assert the same call shapes.

**After the fix: 1076 passed, 6 skipped, 0 failures.** `makemigrations --check` → "No changes detected". Frontend: `tsc -b` clean, 34/34, remote builds.

## Reviewed

Both halves of the gate on `origin/dev...HEAD`. `codex review` surfaced nothing new about the merge. The fresh-context pass came back "safe to push" and independently re-derived the ADR blob equality, the merge-tree equivalence and the `create=True` mechanism rather than taking them on trust.

It also confirmed the things worth confirming before this lands in `dev`: `TRIALS_DB_TOLERATE_MISSING_COLUMNS` is off by default and refused on the primary database three independent ways (alias absent, same normalized host/port/name as `default`, and a config-independent `django_migrations` backstop), introspection failure fails **closed**, the PHR `alg` handling is an explicit allowlist rather than an `else` with `algorithms=["RS256"]` re-pinned at decode, `exp` is required-not-just-verified, and with `PHR_BASE_URL` unset the new default chain returns `None` in ~4 ms **with no socket opened**.

## Follow-ups (none blocking, all pre-existing in the stack — your call)

1. **`/healthz` inherits DRF throttling** — `throttle_classes == [AnonRateThrottle, UserRateThrottle, ScopedRateThrottle]`, `anon: 60/minute`, no override and no test. Probers behind one NAT, or an interval under a second, get 429 and the load balancer pulls a healthy instance. One line: `@throttle_classes([])`.
2. **`PHR_JWKS_MIN_REFRESH_INTERVAL=0` removes the floor entirely**, restoring one blocking outbound fetch per unknown `kid`. The introspection budget guards exactly this with `max(interval, 1)` at `phr.py:81`; the asymmetry looks unintended.
3. **Introspection POSTs `json={"token": …}`** — RFC 7662 §2.1 wants form encoding, and there is no client auth. This is the one item genuinely lost by leaving `8027f3c` out. Local-dev only today (the path is gated off when deployed).
4. **Armed tolerance is asymmetric**: `details()` skips a deferred column so the endpoint returns 200 with the attribute simply *absent* — no null, no marker — while `UserToTrialAttrMatcher` reads trial attrs by bare `getattr` and still 500s on the same request. Two policies on one response, and an eligibility table cannot distinguish "corpus has no value" from "trial does not restrict on this". Inert until the flag is set, but worth a line in `docs/setup.md`.
5. **Successful introspection is cached for the process lifetime**, so once the corpus catches up, live workers keep deferring columns that now exist until restart. Documented in the module, missing from the `docs/setup.md` runbook row. Also, `install()` patches 50+ managers with no "armed" log line at all.
6. **`requirements.txt`** — `httpx`, `PyJWT` and `cryptography` are ranges (`cryptography>=42.0` with no upper bound) in an otherwise fully `==`-pinned file, and they are the auth path's libraries.
7. **`deploy-render.yml` is inert on `dev`** (GitHub resolves `workflow_run`/`workflow_dispatch` from the default branch only, and the file is on neither `main` nor `dev` today). Its guards are sound, but the job hard-fails when `vars.RENDER_EXACT_SERVICE_ID` is unset — worth confirming that and `RENDER_API_KEY` exist before it reaches `main`, or every green `main` push goes red.
8. **`Dockerfile` now has a Node 24 stage** building the same artifact `Dockerfile.gcp` builds on Node 20. Redundant rather than harmful — only `docker-compose.yml` uses the plain one and `WHITENOISE_ROOT` falls back correctly — but `docker compose up` on a fresh clone now needs a full `npm ci` + `vite build` first, on a different Node major than anything shipped.

#415 still merges cleanly onto the updated base (`mergeStateStatus: CLEAN`).

PR #416, opened 2026-09-09
docs: federated Find Trials parity plan (gap analysis + phases)

LeoMo42:
## What

A planning doc for bringing the federated EXACT **Find Trials** remote to CancerBot-level UX in both hosts (HT PHR at `/curehub/trials` and the CB-vendored widget), with the remote staying self-contained.

Three gap matrices (search page, detail page, API), the architecture the plan rests on, and five phases sized in days.

Doc only — no runtime change.

## The two decisions it records

**PROMOP owns per-user state** (favorites, saved filters), not EXACT and not each host. EXACT stays stateless, and PROMOP already models the patient↔trial edge as `PatientTrialEnrollment(person, trial_id, nct_id, status)` with CRUD at `/api/v1/trial-enrollments/` — so favorites is one field (`is_favorite`) on an existing table rather than new storage. Cross-checked with codex, which reached the same conclusion independently.

**The Favorites tab intersects with the matcher inside EXACT's queryset.** Bookmark ids travel into the `match` actions as a bounded `trial_ids` list, so filtering, sorting, pagination, and `itemsTotalCount` stay server-side and correct. Fetching favorites client-side would break all four.

## Three defects found while writing it

All scheduled in phase 0, all small:

- **`?phase=` silently does nothing.** `StudyPreferences` has the field and the queryset filters on it (`by_phase`), but `study_preferences_from_query_params()` never populates it. One line.
- **`?type=favorites` is a 500.** `trials_views.py:191` runs `queryset.filter(favorite=True)`; EXACT's `Trial` model has no `favorite` field and nothing annotates one (CB does, via `add_favorite` — that call has no counterpart here). FieldError.
- **Sorting is unreachable from the inline-payload path.** `/trials/match/` binds to `list`, whose ordering is hard-coded to `-match_score`; only the `search` action honours `?sort=`, and it has no POST alias.

## Deferred

Registered / Register Interest is left out pending Adam — the PROMOP model already supports it (`status ∈ {interested, registered, entered, completed, withdrawn}`), so the open question is product, not technical.

## Review

`codex review --base origin/2omop` — no blocking issues. The load-bearing claims about EXACT's code were re-verified against the source before commit (absence of `add_favorite`, the `graph` action on `TrialsGraphViewSet`, `closestLocationGeoPoint` in the serializer).

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ Patient profiles are passed inline with each API request. They can also be retri
 | [docs/evaluator.md](docs/evaluator.md) | Evaluating EXACT results against ground truth |
 | [docs/setup.md](docs/setup.md) | Local development setup |
 | [docs/api.md](docs/api.md) | REST API reference |
+| [docs/federated-ui-parity-plan.md](docs/federated-ui-parity-plan.md) | Federated Find Trials remote: gap analysis vs CancerBot and phased plan |
 
 ## Running trial search for patients
 
--- a/docs/federated-ui-parity-plan.md
+++ b/docs/federated-ui-parity-plan.md
@@ -1,0 +1,169 @@
+# Find Trials: CancerBot Parity Plan
+
+**EXACT federated remote — gap analysis and phased plan**
+
+What diverges between CancerBot `ui.v2` and the federated EXACT remote — in layout, UX, and API — and the order in which to close it, keeping the remote self-contained for both hosts (HT PHR and CB).
+
+*9 September 2026 · Sources: CB `ui.v2/client`, EXACT `frontend/src/federation`, EXACT/CB `trials/api`, `ht-phr/frontend`*
+
+---
+
+## 1. Decisions from the interview
+
+| Question | Decision | Consequence |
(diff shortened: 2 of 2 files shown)


LeoMo42:
Issues filed against the phases in this plan:

| Phase | Issue |
|---|---|
| 0 — fix the three defects the analysis surfaced | #417 |
| 1 — core parity: tabs, sort, pagination, filter panel | #418 |
| 2 — favorites and saved filters | #419 + healthkey-ai/promop#1142 (storage) |
| 3 — map, CSV export, knowledge graph, suitability weights | #420 |
| 4 — inline editing of "Your Value" | #421 |

Registered / Register Interest has no issue on purpose — it is a product question for Adam, not scheduled work.

LeoMo42:
Re-basing this onto the new `cb-like-trials` integration branch, which forks from `dev` — see the replacement PR. Closing rather than re-targeting because the diff against a `dev`-based branch would otherwise carry all 221 commits that separate `2omop` from `dev`.

Issue #438, opened 2026-09-11
The federated remote has no stable notion of "which person is this" for the inline-payload host


LeoMo42:
Surfaced while reviewing the filter-persistence work (#419).

## The problem

`TrialMatches` needs a patient identity for three different jobs, and has one string to do all three:

```ts
const stateKey = `${personId ?? ""}|${patientInfoKey ?? ""}`;
```

It was introduced as a **cache key**, where maximum discrimination is exactly right: any difference in either prop must be a different key, or one reader is served another's bookmarks. It is now also used as an **ownership key** — whose panel edits are these, whose write is this, whose failure message is this — where the question is different: *is this the same person?*

For the `personId` path the two coincide. For the inline-payload path they do not: `patientInfoKey` is `JSON.stringify(patientInfo)`, so any change to the payload's **content** for the same person — the host refetched a profile that has since been edited, a field was added to the payload, a value was normalised — reads as a different patient.

## What that costs today

- The panel drops the reader's in-session filter edits and reverts to their stored set.
- A save still inside its 500ms debounce is dropped instead of written (`hooks.ts`, the `write.key !== keyRef.current` guard).
- Nothing on screen says either happened.

It fails in the safe direction — it never writes one reader's filters into another's row — and the window is small. But it is a silent loss, and the same identity now governs the bookmark/registration write record too (`writes.key`).

## Why it was not just fixed

Every alternative is a guess about a host's opaque payload:

- Prefer `personId` when present, fall back to the payload — helps only hosts that already have the stable identifier.
- Look for `id` / `personId` / `patientId` inside `patientInfo` — inventing a contract the payload does not have.
- Compare payloads on a subset of fields — picking the subset is the same guess.

The honest fix is for the remote to be **told**: an optional `patientKey` prop a host sets to whatever it considers the identity of the person on screen, defaulting to today's behaviour when absent. Cheap for HT PHR (the person id) and for CB (its patient record id), and it makes the cache key and the ownership key separable — they can stay different strings, which is the point.

## Where

- `frontend/src/federation/TrialMatches.tsx` — `stateKey`, `touched` / `filtersAreMine`, `writes.key`
- `frontend/src/federation/hooks.ts` — `useFilterPersistence`'s key guard
- `frontend/src/federation/types.ts` — the new prop

Issue #437, opened 2026-09-11
Persist the trial type filter, once the panel can check it against the disease


LeoMo42:
The federated filter panel now saves the patient's filters to PROMOP (#419). Ten of the eleven panel fields are persisted; `trialType` is deliberately **not**.

## Why it is left out

A trial type is scoped to the disease. The options come from `/form-settings/?disease=`, and `by_trial_type` has no leniency for a value that is not in them — it matches nothing, so the list comes back empty from a control that is rendering blank.

The component already treats a type as belonging to the patient it was picked for (`trialTypeOwner` in `TrialMatches.tsx`): a pick made under one patient goes stale under the next. A **stored** type would arrive unclaimed — nobody picked it this session — so nothing would ever make it stale. Same patient, disease changed between sessions: the stored type is applied, has no matching option, and the search silently returns nothing.

## What it would take

CB solves this with an effect that drops a type once the options for the settled purposes come back without it (`ui.v2/client/components/trials/TrialFiltersPanel.tsx`, the `typeOptions` effect). The federated panel would need the same: keep the stored value out of the request until the disease's options have loaded, then apply it only if it is among them — and drop it, rather than leave an invisible filter running, if it is not.

Two details CB got right and worth copying: judge against `isFetching`, not `isPending` (the previous disease's options stay on screen across the round trip and must not be what a type is judged against), and treat an empty options list as "the request failed", not as "no such type" — dropping a filter the reader set on the strength of a blank response is worse than keeping it.

## Where

- `frontend/src/federation/filters.ts` — `PERSISTED_FIELDS`, and the comment there explaining the exclusion
- `frontend/src/federation/TrialMatches.tsx` — `trialTypeOwner` / `trialTypeIsStale`
- `frontend/src/federation/FilterPanel.tsx` — already holds the options query

Until then a reader's trial type lasts the session and no longer, which is honest: it is better than an empty result set nothing on screen explains.

PR #442, opened 2026-09-11
feat(federation): the filter panel, saved where the patient left it (#419)

### LINKED ISSUE #419: Phase 2 — federated Find Trials: favorites and saved filters
Third phase of the federated **Find Trials** parity plan (`docs/federated-ui-parity-plan.md`, PR #416). ~3–4 days. Depends on #417; pairs with healthkey-ai/promop#1142, which does the storage.

## The decision this rests on

PROMOP owns the per-user state; EXACT stays stateless. PROMOP already models the patient↔trial edge as `PatientTrialEnrollment` with CRUD at `/api/v1/trial-enrollments/`, so favorites is one field on an existing table. Cross-checked with codex, which reached the same conclusion independently.

## EXACT side

- [ ] **`trial_ids` in the body of the `match` actions** — validated and capped (≤500). The Favorites tab is then an ordinary search narrowed by that list, so filtering, sorting, pagination and `itemsTotalCount` all stay server-side and correct.
- [ ] Retire the phase-0 placeholder 400 on `?type=favorites` in favour of this path.

The cap is not optional: an unbounded id list expands into thousands of WHERE clauses. Over the cap the UI degrades with an explicit message rather than silently truncating.

## Remote side (`frontend/src/federation/`)

- [ ] **State adapter**: an optional `state` prop — `{ listFavorites, toggleFavorite, getPreferences, savePreferences, resetPreferences }` — with a default PROMOP REST implementation built from a `promopClient` prop plus `personId`. HealthTree PHR authenticates via an OAuth mint and CB via the in-process plugin, so the transport has to be injectable. With no adapter the remote degrades to localStorage and the dev harness keeps working.
- [ ] Bookmark control on the card and in the detail view; Favorites tab with its count.
- [ ] Filters read and written through the adapter, with debounce **and a write queue**. CB's queue (one PATCH in flight, the next queued behind it, a generation counter so Reset retires a write already on the wire) exists because ticking two filters in one frame really does land out of order — port the behaviour, not just the look.

## Host side

- [ ] Pass `promopClient` into the remote from HealthTree PHR.
- [ ] CB adapter over the in-process promop plugin.

## Out of scope

Registered / Register Interest — deferred pending Adam. The PROMOP model already supports it (`status ∈ {interested, registered, entered, completed, withdrawn}`), which is exactly why favorites goes in as a separate boolean rather than a sixth status.

LeoMo42:
The last of phase 2 in `docs/federated-ui-parity-plan.md`. Filters a patient sets now survive the session: read from PROMOP on mount, applied to the first search, written back as they change, cleared through the adapter's own reset. With #435 and #436 merged, this closes #419.

## The property everything follows from

**A save is absolute** — it replaces the stored set whole. That one fact produced almost every finding below, in both directions.

- Writing one field while the rest are unknown **discards every filter the patient had** — on the strength of a panel showing the defaults precisely *because* they could not be read. Saving is inert until the read succeeds, the panel is not editable until then, and the banner says so: *changes you make now apply to this search but won't be saved*. Reset is gated the same way, not because clearing needs to know what was there, but because that banner would otherwise be a lie in the most expensive direction.
- It also means the write queue holds **one write per patient**: a later write says everything an earlier one would have — but only within one patient's preferences. B's filters say nothing about A's, and with a single slot B's next edit swallowed A's queued reset.
- And it means reset goes through that queue too. Unserialised, a reset and a save can be in flight together, and the server applies them in whichever order they arrive.

## Key groups, adapter delivers

These are not the same thing, and conflating them was three separate bugs. A host that builds its adapter inline — `state={createPromopState({client, personId})}`, the obvious way to write it — hands over a **new object on every render**, all of them the same person. So the debounce and the queue are grouped by the component's patient key, while each write travels on the **adapter it was made with** — which is bound to one person, so it lands where it belongs no matter what the props do afterwards.

Judged against the component's *current* patient instead: a change made a moment before the host moved on was dropped, and so was a reset queued behind an in-flight save — leaving the reader holding exactly what they had cleared.

## What is stored, and what is trusted coming back

Saved: the ten fields the panel actually renders. Not `country` (derived from the patient), not `type`/`sort` (the tab bar and the sort control own those), not the host's own `initialFilters` — stored, they outrank the host next mount and outlive a host that stopped sending them. Not `trialType`, which is disease-scoped: a type chosen under one disease has no matching option under another and `by_trial_type` has no leniency for a value that is not there, so a stored one produces an empty list from a control rendering blank. Persisting it needs the panel to check it against the disease's options first — **#437**.

Read back, the payload is treated as a claim, not as a `FilterState`. PROMOP checks only that it is an object, and this remote is not its only writer: unknown keys, wrong types, over-long strings, a distance that is not a usable radius, a unit qualifying nothing, and a `type` or `sort` smuggled in as a filter are all dropped.

## Findings worth naming

| | |
|---|---|
| **The read gated the whole list for 7 seconds** | "A failed read is not a reason to wait" was true only of the suite's retry-free client; under a host's own, `status` stays `pending` through three retries with backoff. A *disabled* query is pending for ever too, so an adapter that cannot be asked at all stalled it permanently. |
| **The first edit dropped the other saved filters for 400ms** | The debounce is seeded at mount, before the stored filters exist. The Sponsor box read "Acme", the badge counted it, and the request did not carry it. |
| **A stored value the host happened to agree with was deleted** | And a saved radius lost its *unit* the same way: 50 miles came back as 50 km — a different set of trials, nothing on screen to say why. |
| **A save built during a reset rebuilt what the reset was clearing** | The cached stored set is the pre-reset one until the write lands, and that cache is what tells a save to keep a value the host agrees with. |
| **"No trials found" during the read** | A disabled query reports `isLoading` false, so nothing said the list was loading and the empty state fired instead. Every reader with saved filters saw it. |
| **A save failure outlived its patient**; **refs written during render**; **an adapter that throws synchronously** deadlocked the queue for ever (`.finally` never ran, `inFlight` stayed true) | |

Three pieces turned out to be dead and are gone: a `canSave` flag nothing read, a `generation` counter no path could reach with a stale value, and two lines in reset that cleared what was already null. The debounce's timer and payload are now one ref, so "there is a timer" and "there is a payload" stop being the same fact kept in two places.

## Judgment calls, stated

- **`listAdvancedEnrollments`-style capability check is all-or-nothing.** An adapter must implement read, write *and* clear or the feature is off: with only some, it does not degrade into a smaller feature but into a wrong one — filters that load and then cannot be changed.
- **A value the reader selects that happens to equal the host's is stored as theirs**, once they have any stored value for that field. Codex would rather it were dropped; the alternative makes choosing the value the host already uses mean "delete my preference", which is a surprising thing for a dropdown to do and cannot be told apart on screen from choosing it on purpose.
- **One limitation documented rather than fixed**: the identity these rules group on is `stateKey`, chosen for maximum discrimination as a *cache* key, so for the inline-payload host a change to the payload's content for the same person reads as a different patient and discards unfinished panel edits. The write half is no longer affected (the adapter routes it). **#438** proposes the real fix — a `patientKey` prop, so the remote is told rather than guessing.

## Tests

241, `tsc -b --force` clean, app and remote builds green. **56 mutations, all caught** — including "trust the cache during a reset", "one queue slot for everyone", "drop another patient's debounce", "group the queue by adapter instead of key", "call the adapter directly instead of through a promise", and "save without having read".

Reviewed by two fresh-context passes and seven `codex review` passes; every finding is either fixed above or listed under judgment calls.

### DIFF
--- a/frontend/src/federation/TrialMatches.test.tsx
+++ b/frontend/src/federation/TrialMatches.test.tsx
@@ -403,7 +403,20 @@ describe("a host's initialFilters.trialType", () => {
       expect(screen.getByRole("button", { name: /Filters \(2\)/ })).toBeInTheDocument(),
     );
 
+    // The whole panel belongs to the patient it was filled in for, not just
+    // the trial type: a new patient gets the host's scope back, and (with a
+    // state adapter) their own saved filters over it. So nothing of the
+    // previous reader's is counted here.
     view.rerender(ui({ disease: "multiple myeloma", ref: "B" }));
+    await waitFor(() =>
+      expect(screen.getByRole("button", { name: /Filter Results/ })).toBeInTheDocument(),
+    );
+
+    // Re-armed under the new patient, so Reset has something to clear.
+    await userEvent.selectOptions(
+      await screen.findByLabelText("Recruitment status"),
+      "RECRUITING",
+    );
     await waitFor(() =>
       expect(screen.getByRole("button", { name: /Filters \(1\)/ })).toBeInTheDocument(),
     );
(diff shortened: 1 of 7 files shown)


LeoMo42:
Superseded by #439, which landed the same feature 24 minutes earlier and is already in `cb-like-trials`. Closing rather than resolving the conflicts: on this branch's base, "resolve" would mean picking between two implementations file by file, and the likely outcome is #439 partly reverted with nobody noticing.

Not wasted, though — three things this branch fixed are still live in the merged implementation, and I checked each against `cb-like-trials` itself rather than against my own memory of it:

| probe against the merged code | result |
|---|---|
| how many searches run on mount with saved filters | **two** — `[null, "Acme"]`: the first goes out before the saved set arrives, the second after it |
| does the panel ever show a filter the request does not carry | yes — Sponsor reads "Acme", the badge counts it, the request has no sponsor (the debounced fields are seeded at mount, before the saved set exists) |
| what a stored value of the wrong type does | `phase: 42` reaches the query parameters unchallenged |

I'll port the fixes onto current `cb-like-trials` as a small PR: `sanitizeStoredFilters` (whitelist + per-field type checks, so a payload written by another client or an older build cannot become an invisible filter), gating the opening search on the saved-set load, and seeding the debounce so a loaded value is not held back for 400ms.

One thing for whoever owns `preferences.ts` to decide, because it is a premise rather than a defect I should unilaterally change: `adapterPreferences` is written for an endpoint that merges a partial update, and sends explicit `null`s for keys that have gone away. PROMOP's `upsert` merges at the FIELD level — a body without `preferences` leaves it alone — but `preferences` itself is one JSON column, so a body that contains it replaces the whole dict (`TrialSearchPreferencesSerializer` + `serializer.save()`). Two consequences: the nulls are stored verbatim in the row (invisible, since `get` strips them, but they accumulate), and `forget()` does not make the next save additive the way its comment intends — under a wholesale replace, a save after a skipped load still overwrites the stored set with a subset, which is the exact hazard it was written to prevent.

Details in the branch history if useful; the branch stays pushed for reference. Closes nothing on its own — #419 is delivered by #439.

Issue #421, opened 2026-09-09
Phase 4 — federated Find Trials: inline editing of "Your Value" on the detail page


LeoMo42:
Fifth phase of the federated **Find Trials** parity plan (`docs/federated-ui-parity-plan.md`, PR #416). ~5–7 days, and the one phase that should be decided on its own before it starts. Depends on #418.

This closes CancerBot's main loop: the patient sees a gap in the eligibility table, fills it in, and the match recomputes. The remote's detail page is read-only today.

## Write path — the constraint that shapes the whole phase

In PROMOP, `PatientRecord` is a **read-only projection**, re-derived by a signal from the OMOP tables. Writing to it directly is not a supported path. The two sanctioned paths are granular CRUD on the OMOP tables (`/api/v1/measurements/`, `/conditions/`, …) and `PATCH /api/v1/patient-records/{person_id}/`, which does *not* write the projection but translates each field into the matching OMOP table write and then triggers the re-derivation.

Inline editing should use the second. That makes the first task an audit, not code.

## Work

- [ ] **Mapping audit first**: for every editable eligibility field, does a write path into OMOP actually exist via `PATCH /patient-records/{id}/`? Produce the table. Fields without a mapping stay read-only — a control that appears to save into nothing is worse than no control.
- [ ] Port `FieldEditControl` / `SubformDialog` from CB `ui.v2/client/components/details/`.
- [ ] Port the autosave queue: per-field batching into one PATCH, one request in flight at a time, edits made during a flight drained afterwards, flush on unmount, error toast, and the optimistic value retired whether the save succeeded or failed. CB earned each of those the hard way; copy the behaviour, not just the markup.
- [ ] `subform_details` — EXACT already sends them and the remote ignores them. At minimum a view-only dialog, ideally the editable one.
- [ ] After a successful write, invalidate the match query so scores and row statuses recompute.

## Open question

Whether patient-authored values should be provenance-tagged differently from EHR-ingested ones. PROMOP has provenance machinery; this phase is the first time the remote writes anything, so it is the moment to decide.

LeoMo42:
Base: this lands on the `cb-like-trials` integration branch, which forks from `dev` (not `2omop`). On that line the trial queryset is `trials/querysets/trial.py` rather than the packaged `matcher_app/exact_matching/`, so backend edits are written against `dev`'s layout.

LeoMo42:
## Mapping audit — the first task, done

[...]

Issue #420, opened 2026-09-09
Phase 3 — federated Find Trials: map, CSV export, knowledge graph, suitability weights


LeoMo42:
Fourth phase of the federated **Find Trials** parity plan (`docs/federated-ui-parity-plan.md`, PR #416). The CancerBot widgets that sit around the trial list. ~4–5 days. Depends on #418.

Most of this is UI over endpoints that already exist — worth noting, because the phase looks heavier than it is.

## Work

- [ ] **`/trials/export/`** — port CB's streaming-CSV action (`type=eligible_and_potential|favorites|my_trials`, `sort=`), minus the persisted-preferences coupling.
- [ ] **Map view**: List/Map toggle, pins, sticky panel, expand. The data is already served — `closestLocationGeoPoint` is on every list row. The Google Maps key arrives from the host as a prop; the remote carries no key of its own.
- [ ] **Knowledge graph**: the backend already exists as `/trials-graph/graph/` (`TrialsGraphViewSet.graph`), so this is UI only.
- [ ] **Suitability weights**: `benefitWeight`, `patientBurdenWeight`, `riskWeight`, `distancePenaltyWeight` are already accepted by `get_queryset` and default to 25.0 each. Needs the control plus storage in the same `TrialSearchPreferences` from phase 2.
- [ ] **`highRiskMclCriteriaBreakdown` panel** — EXACT returns it from `retrieve` today and the remote drops it on the floor.
- [ ] **Port `matchingType: not_evaluated`** (CB #4850) into EXACT and render the third state. Today a criterion the matcher judged uninterpretable (`sane_range`) comes back as `matched` and renders a green tick against a requirement nobody checked. That is the one item here that is a correctness fix rather than a feature, so it can be pulled forward if the rest slips.

LeoMo42:
Base: this lands on the `cb-like-trials` integration branch, which forks from `dev` (not `2omop`). On that line the trial queryset is `trials/querysets/trial.py` rather than the packaged `matcher_app/exact_matching/`, so backend edits are written against `dev`'s layout.

PR #443, opened 2026-09-11
fix(federation): three things the saved filters get wrong on their way in (#419)

LeoMo42:
On top of #439, which landed saved filters. This is the salvage from #442 — the same feature built independently, closed rather than merged because resolving it on its base would have half-reverted #439 invisibly. Three defects survived the comparison, and each was **measured against `cb-like-trials` itself** with a throwaway probe before anything was touched:

| probe on the merged tree | result |
|---|---|
| how many searches run on mount with saved filters | **two** — `[null, "Acme"]`: the first before the saved set arrives, the second after |
| does the panel ever show a filter the request does not carry | yes — Sponsor reads "Acme", the badge counts it, the request has none |
| what a stored value of the wrong type does | `phase: 42` reaches the query parameters unchallenged |

## The fixes

**One search, not two.** `useSavedFilters` now reports `pending` and the trials query waits on it — **capped at 300ms**, because the list must not be hostage to the preferences service. A read that is merely slow gives the reader an unfiltered list after the grace period and a corrected one when it arrives, which is what happened every time before the gate existed; a read that *fails* releases it at once. #439's three tests asserting the list starts without waiting still pass unchanged.

**The panel and the request agree.** `useDebounced` gained a pass-through mode: a value from the saved set is not a keystroke, and holding it behind the 400ms only produces a request that disagrees with the controls. Gating alone would not have fixed this — the first request would still have carried the stale held values.

**The stored set is a claim, not a `FilterState`.** `sanitizeStoredFilters` drops unknown keys, wrong types, empty and over-long strings, a `type` or `sort` smuggled in as a filter, and a radius whose unit cannot be read. That last one is a judgment: keeping the number looks harmless, but `by_distance` treats anything other than `miles` as kilometres, so a stored 50 miles would quietly become 50 km — from a value we had just admitted we could not parse.

## What review found in the fixes themselves

Three rounds, and the first finding is the one I had written a warning about in my own commit message and then not acted on:

- **"No trials found", on every mount, before anything was asked.** A disabled react-query query reports `isLoading` false, so gating the search silenced the loading line and let the empty state take its place. The gate now appears in all three conditions, the way `waitingForIds` already was two lines above it.
- **The gate was defeated under StrictMode** — which is every entry point in this repo, so the developer verifying the fix in the dev harness would have watched it fail. The mount effect is double-invoked for the *same* writer, and ownership cannot tell the two reads apart: the abandoned one opened the gate belonging to the live one.
- **The gate belonged to the patient and should belong to the writer.** The two come apart when a host gains an adapter for the same patient: the panel was on `localStorage` a moment ago, the key never changed, and PROMOP's slower read was still in flight.
- **It was also raised in an effect**, a render after the query observer had been handed the new key.
- **`typing` was one-way**, so it stayed true across a switch and the next patient's saved set went through the delay anyway. Making it per-patient then broke the switch *back*: returning to someone whose panel had been typed in turned it on again. What matters is whether the panel has been touched since the last switch.
- **A units-only preference could not be reverted.** `userOwnedFilters` emits `distanceUnits` only when it differs from the baseline, and is the one saved field that never consulted the sticky `owned` set — invisible while the sanitizer was dropping standalone units, and exposed the moment it stopped. On the merge-only localStorage transport the reader could not get back to the host's kilometres except through Reset. Now it honours `owned` like every other field.
- **`validatedOnly: false` was dropped**, so unchecking a box the host had ticked did not persist — the same shape of action as the units change, decided the opposite way. A boolean is stored as a boolean.

Three of my own tests turned out to assert nothing (two checked that a stored `type`/`sort`/unknown key does not reach the request, which `filterStateToParams` makes structurally true whatever the sanitizer does), and two more described scenarios they did not construct. They are gone; the same properties are pinned where they can fail. The sanitizer's field allowlist is now pinned **by name** — six of the fourteen were carried by no test at all, and a missing name there does not merely ignore a filter, it deletes it on the next save.

## Coordination

Agreed with the author of #439, who reviewed the overlap and asked for exactly these three ("I'd rather have your sanitizer than write a second one"). Phase 3 (#420) is theirs; #421 is mine.

Separately, and **not** in this PR because it is their module's premise to change: `adapterPreferences` is written for an endpoint that merges a partial update. PROMOP merges at the *field* level, but `preferences` is one JSON column, so a body containing it replaces the whole dict — which made `forget()` the opposite of what its comment intended. They confirmed it independently and are fixing it on `fix/419-wholesale-replace`. A promop-side issue about the contract is still to be filed; GitHub is currently refusing content creation on this account.

One thing worth carrying forward for anyone gating a query again: a disabled query reports `isLoading` false, so the empty state fires in place of the loading line unless the gate is in both.

## Tests

267, `tsc -b --force` clean, app and remote builds green. **28 mutations, all caught** — including "cancelled read opens the gate", "gate not re-armed for a new writer", "units revert not owned", "loading line ignores the gate", and the six-name allowlist narrowing.

One of #439's tests was adjusted rather than rewritten: the tab-count assertion now awaits the badge instead of reading it synchronously, because the first response is no longer in hand by the time the tab bar renders. Same assertion.

### DIFF
--- a/frontend/src/federation/FilterPanel.tsx
+++ b/frontend/src/federation/FilterPanel.tsx
@@ -10,7 +10,7 @@
 // without saying which. Single-select until that parser is ported — #428.
 import type { AxiosInstance } from "axios";
 
-import { DISTANCE_UNITS, isActiveDistance } from "./filters";
+import { DISTANCE_UNITS, LAST_UPDATE_OPTIONS, isActiveDistance } from "./filters";
 import { useFormSettings } from "./hooks";
 import type { FilterState } from "./types";
 
@@ -78,6 +78,23 @@ function SelectFilter({
   // value; when one is missing we supply the placeholder ourselves, so the
   // control always offers a way back to no filter.
   const hasEmpty = options.some((option) => option.value === "");
+  // A value the option list does not contain — one stored by another
+  // client, one from a host's `initialFilters`, one whose list belongs to
+  // a different disease — otherwise collapses this control to its
+  // placeholder: the reader is shown "Any" over a list the value is
+  // narrowing, with the badge counting a filter they can neither see nor
+  // clear from here. Showing it, even under its raw code, is what makes it
+  // clearable. (Deciding whether a value is legal is #444; this is the
+  // half that needs no answer from the server.)
+  //
+  // It also covers a legal value while the catalog is still on its way, so
+  // a saved filter reads as its own code for that moment and then as its
+  // label. That is the honest order: the code is what the request is
+  // carrying either way.
+  const shown =
+    value && !options.some((option) => option.value === value)
+      ? [...options, { value, label: value }]
+      : options;
   return (
     <Field label={label}>
       <select
@@ -86,8 +103,11 @@ function SelectFilter({
         onChange={(e) => onChange(e.target.value || undefined)}
       >
         {hasEmpty ? null : <option value="">Any</option>}
-        {options.map((option) => (
-          <option key={option.value || "__any__"} value={option.value}>
+        {/* Keys are prefixed so the key space is the values themselves:
+            keyed on `option.value || "__any__"`, a stored value of literally
+            `"__any__"` collided with the server's own empty-value entry. */}
+        {shown.map((option) => (
+          <option key={`opt:${option.value}`} value={option.value}>
             {option.label}
           </option>
         ))}
@@ -172,12 +192,7 @@ export function FilterPanel({
         <SelectFilter
           label="Updated within"
           value={filters.lastUpdate}
-          options={[
-            { value: "1", label: "the last year" },
-            { value: "2", label: "the last 2 years" },
-            { value: "3", label: "the last 3 years" },
-            { value: "5", label: "the last 5 years" },
-          ]}
+          options={[...LAST_UPDATE_OPTIONS]}
           onChange={(v) => set({ lastUpdate: v })}
         />
 
--- a/frontend/src/federation/TrialMatches.test.tsx
+++ b/frontend/src/federation/TrialMatches.test.tsx
@@ -13,6 +13,7 @@
 // They assert on the request log rather than on pixels: which request went
 // out, and when, is the thing that was wrong.
 import { QueryClientProvider, QueryClient } from "@tanstack/react-query";
+import { StrictMode } from "react";
 import { render, screen, waitFor } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { beforeEach, describe, expect, it, vi } from "vitest";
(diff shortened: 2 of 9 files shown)


LeoMo42:
**Sequencing note for whoever merges.** The author of #439 has `fix/419-wholesale-replace` ready and unpushed, which rewrites `preferences.ts` substantially — `forget()` removed from the interface and both transports, `adapterPreferences` merging over what the server holds instead of sending nulls, a generation counter, and a changed `useSavedFilters` callback signature.

We agreed this one goes first if the order is ours to pick: it is open, reviewed and smaller against the current base, and they would rather own the resolution afterwards than hand over a diff that deletes a method someone just built on. **But the merge order is yours, not ours** — if theirs lands first instead, no harm: they will resolve in that direction and I will rebase this branch and re-run its gate.

Checked so the resolution is cheap either way: this PR does **not** touch `forget()` — the only mentions in the diff are context lines. What it does add to the same promise chain is the `sanitizeStoredFilters` call in the `.then`, a `.finally` that releases the gate, and `pending` on the hook's returned object, which is where their `onError` line will also want to go.

One thing for them rather than for the merge: my gate is owned by the writer object and their generation counter is a second identity for nearly the same question. One of the two should win in the merged result — two notions of "which attempt is current" in one hook is how the bugs above got in.

LeoMo42:
Force-pushed. The branch is the same three fixes, but the review found enough in them that the diff roughly doubled — and most of what it found was in **my corrections**, not in the original three.

Ten rounds: a fresh-context read and a `codex review` on each, the codex half on `gpt-5.6-sol` from round five onward at Leonid's request. The newer model found on its first pass something three passes of the previous one had not: the gate stopped the *request* and not the *rows*, so react-query went on serving the previous patient's answer under "Loading trials…".

**What the rounds actually found**, because the pattern matters more than the list:

| round | defect |
|---|---|
| 2 | "No trials found" on every mount, before anything was asked — a disabled query reports `isLoading` false |
| 3 | the gate defeated under StrictMode, which is every entry point in this repo |
| 4 | a filter longer than 200 characters written by this client, refused on the way back, then **deleted** from the row |
| 5 | the gate held the request and not the rows, total or badges |
| 6 | it then held the pager too — unmounting the button being clicked, dropping keyboard focus on every page turn |
| 7 | so the pager was let through the whole window, and painted the **previous view's** page count, clickable, 404-ing |
| 8 | the view it checks itself against did not include the patient |
| 9 | widening `lastUpdate` to a range reintroduced round 8's own defect from the other side: a value with no `<option>` collapses the control to "Any" while the request carries it |

Rounds 6 through 9 are all one shape: **a condition applied to one consumer and not to its siblings.** Request without rows, rows without pager, pager without patient. Each fix was correct and each was incomplete in the same way, which is why the code now names the questions once — `hasPatient`, `staleResponse`, `staleRows`, `countIsForThisView` — instead of spelling each out where it is needed.

Two changes reach further than the three fixes, both because a fix of mine made them necessary:

- **`SelectFilter` now shows a value it has no option for**, under its raw code if that is all there is. Widening `lastUpdate` made unknown-but-legal values reachable, and a control that silently reads "Any" over a list it is narrowing is the half of #444 that needs no answer from the server. Every select gets it, not just that one.
- **`userOwnedFilters` consults `owned` for `distanceUnits`** — the one saved field that did not, invisible until the sanitizer stopped dropping standalone units. Without it a reader could not get back to the host's kilometres except through Reset.

281 tests, **62 mutations, all caught**. `tsc -b --force` clean; app and remote builds green. Three of #439's tests were adjusted rather than rewritten, each noted in the commit message.

Still worth reading as the one judgment call rather than a fix: `lastUpdate` is bounded at 2000 years because that is where the backend's own arithmetic overflows (`(datetime.now() - datetime.min).days // 365`, a little past 2027 today and rising by one a year) — not at the four values the panel offers, because refusing a stored value the backend handles gets it nulled out of the PROMOP row by the next save.

Issue #158, opened 2026-06-12
[MEDIUM] graph_view n param unvalidated + serializes missing patient -> 500


LeoMo42:
**Severity:** MEDIUM · **Source:** Codex

**Where:** `trials/api/graph_view.py:70,105`

**Problem:** `n=int(...)` is unvalidated and `GraphPatientInfoSerializer(patient_info)` is called even when no patient was resolved; a bad `n` or missing patient context produces 500s on an authenticated API.

**Fix:** Validate `n` with a serializer and require patient context for graph responses, returning 400 with a stable error.

---
_Filed from the 2026-06-12 platform security/arch audit (8-agent Claude review + Codex second opinion). Full report: `platform-review-2026-06-12.md`._

LeoMo42:
Resolved by PR #162 (merged to main): graph endpoint returns 400 (not 500) for a non-numeric n param and when no patient context resolves.

Issue #150, opened 2026-06-12
[CRITICAL] No object-level authz on ?person_id= patient path (PHI IDOR)


LeoMo42:
**Severity:** CRITICAL · **Source:** Codex (Claude rated HIGH) — independent agreement on the issue

**Where:** `trials/services/patient_info/resolve.py:15,62-105`; `ctomop_client.py:85`; `trials/api/graph_view.py:105`; `patient_info_serializers.py:18`

**Problem:** When a request carries `?person_id=` (or person_id in body), `resolve_patient_info` fetches that patient's full clinical profile from CTOMOP using a static service token, with NO binding between the authenticated Identity and the requested person_id. The graph endpoint returns serialized patient fields. Any authenticated caller can enumerate `person_id` and receive other patients' clinical attributes. Documented as design pin #108 relying entirely on CTOMOP enforcing row-level authz (which it doesn't for this path).

**Fix:** Remove the public `person_id` lookup until scoped, or forward caller identity to CTOMOP and enforce patient/tenant authz there. Add object-permission tests proving user A can't fetch user B's patient. Until #101's identity flow lands, gate behind a feature flag / service-token-only callers.

---
_Filed from the 2026-06-12 platform security/arch audit (8-agent Claude review + Codex second opinion). Full report: `platform-review-2026-06-12.md`._

LeoMo42:
Cross-service angle (2026-06-13 microservices-interaction audit, finding I1): this `?person_id=` IDOR is the **entry point of a confused-deputy chain** — exact forwards an unbound `person_id` to ctomop under the *static shared service token*, so ctomop authorizes the trusted-backend principal, not the end user (`can_access_patient` never runs on that read path). Systemic root: the end-user identity is **not propagated across the hop**. Fix this together with ctomop #134/#135. See `microservices-interaction-audit-2026-06-13.md` (I1).

LeoMo42:
Duplicate root cause as #108 ("Bind CTOMOP person_id resolver to authenticated caller"). #108 carries the design context (EXACT is stateless re: patient data, ties to closed #101/#102 identity-flow) and acceptance criteria, so consolidating work there. Keeping this CRITICAL audit item open for severity tracking until the fix lands; will close both via the same PR. See also infra epic #125.

PR #446, opened 2026-09-11
feat(federation): Explore Trials — the map the list cannot draw

LeoMo42:
Part of #420 — the knowledge graph. Map view and suitability weights remain; CSV export is #445.

The backend has existed since #158 and nothing could call it.

### It was unreachable, and its success path was broken

`/trials-graph/graph/` is GET-only, and the graph refuses to draw without a patient. The two ways to supply one are an inline payload — body-only, since GET-with-body is forbidden by the Fetch spec and dropped by axios — or `?person_id=`, which is gated off outside DEBUG because it would let any authenticated caller read another patient's record (#150/#108). So the endpoint was reachable in local development and essentially nowhere else. A POST alias fixes that, mirroring `search_match`.

With the alias in place it **500s on every successful request**. `PatientInfoSerializer` copies `vars(instance)` wholesale, and `PatientInfo` carries two `cached_property`s — `attributes_service`, which caches a service object, and `geolocation`, which caches the patient itself. A cached property writes its result into the instance dict on first read, so whether the response rendered depended on whether anything had touched the matcher yet. The graph always has. Its existing tests covered only the two 400s, which is exactly why this survived: the success path had never run.

The exclusion is derived from the class rather than a denylist, so a cached property added later is excluded by existing. It drops derived data too — `mutation_genes` and its neighbours are cached properties, and they appeared in the response whenever something had warmed them. That is the trade and it is the right way round: a field present only when an unrelated code path happened to touch it is worse than one consistently absent. A test pins the output to being the same cold and warm.

### The UI

CancerBot draws this with a d3 force simulation across ~4,000 lines. This remote ships six runtime dependencies and the host inherits every one, so a graph library would be the largest thing in the bundle for one screen. The layout is computed instead — which is also better for the reader: a force simulation re-settles on every open, so the picture a patient describes to someone is not the picture they get back.

An attribute is **one node** however many trials ask about it — two cards cannot show a reader that one missing value stands between them and three trials at once. Keyed on the trial's field rather than the patient's, because `age_low_limit` and `age_high_limit` are both `patient_age` and would collapse into one node carrying the first label. Banded by its **worst** answer across the trials on screen; the text under each trial uses **that trial's** buckets instead, or it would print "Not met: ECOG" beneath a trial whose ECOG the patient meets.

Everything in the picture is repeated as text below it — not a fallback, but the copy that works with a screen reader, in a narrow window, and when a label is longer than the space beside its dot. The SVG itself is `aria-hidden` and carries no tab stops: `role="img"` makes its subtree presentational, so clickable groups inside it are nameless tab stops in Chrome and unreachable in WebKit.

The fetch happens when the panel opens and not before — it is a second full matcher run over the same search. It is narrowed by whatever narrows the list, `trial_ids` included, and the control refuses outright when a state tab's ids are unavailable: with no ids the cache key **is** the default tab's, so a merely-disabled query would hand back the whole-corpus map under a tab that says "Favorites".

### Deliberately not ported

The force layout, the tree view, PNG export, dependency sub-trees, compare mode, and concept multi-select highlighting. What is here is the reading the list cannot give; those are elaborations of it. Worth naming rather than discovering.

Also: widening `http_method_names` to accept POST published every inherited POST action under `/trials-graph/` — an untested second copy of the trials API at a URL nobody asked for. They are removed.

### Reviews

Two full gate passes; 13 findings, all fixed.

981 backend tests pass; 254 frontend tests pass; `tsc -b` clean.


LeoMo42:
Rebased onto the moved base (`cb-like-trials` picked up #450 and #451) and re-gated the **combination**, not just this branch — #450 edits `graph_view.py` too.

The delta review found this endpoint **still 500s**, on a field the fix in this PR does not cover. `normalize_patient_info` turns latitude/longitude — or country + postcode — into a GEOS `Point` on `geo_point`. It is a plain attribute, not a `cached_property`, so the rule that drops the service caches leaves it, and `JSONRenderer` raises on it. Every patient with a location gets a 500, and location is a first-class filter here.

Every test on this endpoint, mine and #450's, used a patient with **no** location. That is how it survived: the success path had never run with a real patient, just as it had never run at all before this branch. Two tests now do.

Two more, both artefacts of the combination rather than of either change:

- a therapy row arrived as `patientFieldHasSubform: true` with `patientFieldCanonical: null` — `therapies()` hardcodes `ureadonly` and puts the *trial* attribute in `ufield`, so the flag claims a subform for a row with no patient field to write. Suppressed where there is no field.
- `AttributeNames.get_by_camel_case` rebuilt two lists per call; the graph names every attribute of every trial it draws, so the POST alias put that on the federated path at +38% of wall time. Inverted once at import.

997 backend / 254 frontend pass; both halves of the gate re-run on the merged tree.

Issue #471, opened 2026-09-13
The overwrite register covers one stage; `ctomop_adapter` and `resolve` also write patient values


LeoMo42:
Raised by the fresh-context review of #449, and deliberately not folded into it.

## What #449 answers, and where it stops

`uoverwritten` (and `patientFieldOverwritten` on the graph row) tells a client whether a value it offers a control over will survive the next match. It is computed from a register in `normalize.py`, declared beside the code that writes, and held to it by a test that parses the module.

That register covers **one stage**. Two others write patient values and are invisible to it.

## 1. `ctomop_adapter` — the `?person_id=` path

`trials/services/patient_info/ctomop_adapter.py`:

```python
lines = row.get('therapy_lines_count')
if lines is not None:
    _lines_map = {0: 'None', 1: 'One line', 2: 'Two lines'}
    row['prior_therapy'] = _lines_map.get(lines, 'More than two lines of therapy')
```

and, a few lines earlier, `metastatic_status` from CTOMOP's Positive/Negative.

`prior_therapy` carries **no register entry**, so a row about it answers `uoverwritten: null` — "the reader owns this". On the `person_id` path it does not: whatever PROMOP holds for `therapy_lines_count` overwrites it on every match. An edit is accepted, saved, and reverted, which is exactly the failure #449 exists to name.

`prior_therapy` is not incidental either: `_normalize_therapy_lines` keys the whole therapy-line clearing cascade off it, so this one value decides whether nine other fields are wiped.

## 2. `resolve._coerce_*` — every path

`trials/services/patient_info/resolve.py` drops every non-dict item from `genetic_mutations`, and coerces the therapy list fields similarly. That is a write the reader did not make, to a value they did supply — narrower than case 1, since it removes malformed items rather than replacing good ones, but the same class and the same silence.

## Why it was not fixed with #449

The register's accuracy comes from being declared next to the code it describes and checked by parsing that module. Extending it to two more modules means either three registers, or one register that no single test can hold to its code. That is a design question and wants deciding rather than guessing.

The `person_id` path is also gated off outside DEBUG (#108), so case 1 is not reachable in a deployed environment today. That is a reason to schedule this rather than rush it — not a reason to let the key claim more than it knows, so `normalize.py` now carries a SCOPE note, and this issue is what it points at.

## Shape of a fix

Either:

* a per-module register with the same parse-and-compare test, and `overwrite_note` merging them — honest, and three times the machinery; or
[...]

Issue #470, opened 2026-09-13
Two conditions in `normalize.py` are narrower than they read: a zero coordinate is dropped, and a null creatinine unit is a 500


LeoMo42:
Found while building the overwrite register for #449, which required reading every condition in `normalize.py` closely enough to write it down. Two of them are narrower than they look, in ways nothing on the response explains.

Filed together because they are the same audit and the same module; they are independent fixes.

## 1. A patient on the prime meridian or the equator gets no `geo_point`

`_normalize_geo_point`:

```python
elif pi.longitude and pi.latitude and not (pi.country or pi.postal_code):
    pi.geo_point = Point(pi.longitude, pi.latitude, srid=4326)
```

`0` is a valid coordinate and a falsy number. A patient supplied as `{"latitude": 51.48, "longitude": 0}` — Greenwich — falls through with no `geo_point`, and so does `{"latitude": 0, "longitude": -78.5}` — Quito.

The consequence is not an error. `by_distance` is a no-op without a point, so the radius filter silently does not apply and the reader gets the unrestricted list back under a heading that says otherwise. Same class as #81, where `0` was being read as "missing" for monoclonal protein, and the fix is the same shape: `is None` rather than truthiness.

The prime meridian crosses the UK, France, Spain, Algeria and Ghana; the equator crosses Ecuador, Brazil, Kenya and Indonesia. Narrow, but not hypothetical — and the coordinate path is the one a caller uses precisely when it has no country.

## 2. An explicit `null` for creatinine units is a 500

`EgfrCalculator.call`:

```python
if not scr or not age or not gender:
    return None
...
scr_mg_dl = SerumCreatinineConvertor.call(
    scr, patient_info.serum_creatinine_level_units.lower(), 'mg/dl')
```

The guard covers creatinine, age and gender. It does not cover the units, and the field is dereferenced unguarded two lines later.

`serum_creatinine_level_units` carries `default='MG/DL'`, so this is unreachable from a payload that omits it. It is reachable from one that sends it explicitly as `null`: `_build_in_memory` keeps any key that names a model field, whatever its value, so `PatientInfo(serum_creatinine_level_units=None)` is constructed and `.lower()` raises `AttributeError`.

`normalize_patient_info` runs on every patient-context request, so this is a 500 on the whole endpoint rather than a missing eGFR — the same shape as #391, where an unsupported country made patient normalization raise and took every patient-context endpoint with it.

[...]

PR #472, opened 2026-09-13
feat(api): say which patient values EXACT recomputes

### LINKED ISSUE #449: Nothing tells a client which patient values EXACT recomputes
Surfaced while building the write path for #421, and it blocks the useful half of it.

## The gap

An eligibility row can now name the patient attribute it is about (`upatientField`, #421), and PROMOP's writable-fields descriptor says whether a write would be accepted. Neither answers the question that decides whether a pencil should appear: **does EXACT overwrite this value on the next match?**

`normalize.py` recomputes a substantial set unconditionally — `tnbc_status`, `hr_status`, `tp53_disruption`, `last_treatment`, `treatment_refractory_status`, `mipi_risk`, `mipi_c_risk`, `bulky_disease_criteria`, `high_risk_mcl_criteria` — and more under conditions (`estimated_glomerular_filtration_rate`, `progression`, `stem_cell_transplant_history`, the therapy lines, `supportive_therapies`). A patient who edits one of those sees it accepted, saved, and then silently reverted.

## Why the two flags that look like the answer are not

- **`ureadonly`** is set by `get_field` from the presence of a subform, nothing more. `renal_adequacy_status` and `measurable_disease_imwg` are recomputed and report `false`.
- **`is_computed_value`** in `USER_TO_TRIAL_ATTRS_MAPPING` is a display flag and does not track the pipeline. `_normalize_mcl_derivations` — whose own docstring says the values are "always overwritten here" — sets four fields on consecutive lines and only `high_risk_mcl_criteria` carries it. In the other direction `meets_gelf` and `meets_lugano` carry it and are plain stored `BooleanField`s that nothing derives.

A draft of #421 exposed `is_computed_value` as `ucomputed`, reading it as "EXACT overwrites this". Measured afterwards: 18 of 30 relevant fields disagreed with `normalize.py` in one direction or the other. It would have hidden the only control that can set `meets_gelf`, and offered one over `tnbc_status`, which the next save undoes. The flag was removed rather than shipped.

## What would close it

The overwrite set has to become explicit rather than inferred:

1. Name it in `normalize.py` — a module-level set (or a decorator on each normalizer) listing the fields that function writes, so the register cannot drift from the code that does the writing. Conditional ones want the condition named too: "overwritten when `height` and `weight` are both present" is a different thing to tell a reader than "always".
2. Expose it on the row — one key, alongside `upatientField`.
3. Pin it with a test that walks `normalize.py`'s writes and asserts every one is in the register. Without that, this issue recurs the first time a normalizer gains a field.

Until then an editing client has no way to distinguish a value it may author from one it may only watch, and #421 has to treat the whole set as provisional.

LeoMo42:
Closes #449.

## The question nothing answered

An eligibility row names the patient attribute it is about (`upatientField`), and PROMOP's writable-fields descriptor says whether a write would be accepted. Neither answers the one that decides whether a pencil should appear: **does EXACT overwrite this value on the next match?**

A patient who edits one of those sees it accepted, saved, and silently reverted. "PROMOP will take this write" and "this write will survive" are different questions, and only `normalize.py` knows the second.

## The register

Declared in `normalize.py`, beside the code that does the writing, because a register kept anywhere else is a second copy of the truth. **36 fields — 13 written on every call, 23 under a condition.** The ticket named nine and gestured at the rest.

Several are *clears* rather than computes — `country` is emptied when it names no country EXACT can resolve, `postal_code` when the pair resolves to no point — and a reader whose typo is about to be deleted deserves the same warning as one whose value is about to be recomputed.

The conditions are prose because a client shows them to a reader: *"overwritten when serum creatinine and age are present and non-zero, the gender is M or F, and the creatinine units convert"* is a different thing to be told than *"always"*.

## Why not the two flags that look like the answer

`ureadonly` is set by `get_field` from the presence of a subform, nothing more.

`is_computed_value` is a display flag, and the disagreement is not marginal. Measured across the 133 mapped attributes: it flags **12**, of which **4** are not overwritten by anything, while **16** fields this module does overwrite carry no flag at all — including `tnbc_status`, `hr_status` and every MCL derivation but one. Twenty disagreements either way. Reading that flag would have hidden four real controls and offered sixteen the next save undoes.

## On the row

`uoverwritten`, stamped in the same loop as `upatientField`:

```
null                                       EXACT leaves the value alone
{'when': 'always'}                         rewritten on every match
{'when': 'sometimes', 'condition': '...'}  rewritten under that condition
{'when': 'never-stored'}                   computed on read; nothing to write
```

24 of the 36 reach a row; the other 12 are not in the attribute mapping.

`patientFieldOverwritten` on the graph row too. That endpoint rebuilds each row from a whitelist, and its own comment pointed at this ticket as the answer EXACT did not have — so leaving it out would have left the comment true and the client without it. `null` is also what an un-overwritten field answers, which makes a missing key the hardest kind to notice: every row would read "EXACT leaves this alone".

## The fourth answer, which review had to find

`abnormal_kappa_lambda_ratio` and `meets_meas_or_bone_status` are mapped attributes, reach a row, and are `@property` with **no column and no setter** — recomputed on every read, and dropped by `_build_in_memory` before they reach the instance.

The first draft answered `null` for both: *"this value is the reader's"*, over a field an edit cannot reach at all. Strictly worse than the overwrite case this ticket exists to prevent, and the register alone could never have caught it — the register is about what `normalize.py` **writes**, and these it does not.

That also corrects a claim in the first draft: of the four fields `is_computed_value` flags and nothing overwrites, only `meets_gelf` and `meets_lugano` are plain stored `BooleanField`s. The other two are these.

## The test that makes it a register

`test_overwrite_register.py` parses `normalize.py` and compares its writes against the register, **both ways**. Parsing rather than calling: a normalizer only writes on the branch its inputs select, so exercising `normalize_patient_info` would confirm the register for whatever fixtures happen to exist and miss every branch they do not reach.

Both directions, because they fail differently. A missing entry is the dangerous one — a value presented as the reader's that EXACT overwrites. A stale entry is the quiet one: it hides a control the reader could have had, and nothing would ever surface it.

**The walker had a hole**, and review found it: it matched writes through a local named exactly `pi`, so a helper taking `patient_info` — the spelling the sibling `patient_info_attributes.py` uses — wrote invisibly, as did `setattr(pi, ...)` and `alias = pi`. All three left the register's tests green. The name now comes from each function's own first parameter plus anything bound to one; `setattr` with a literal name is recorded; and a `setattr` with a **computed** name is refused outright, because it cannot be checked by reading and tolerating it would make the guard decoration.

## Mutation-tested

```
remove `tnbc_status` from the register     -> 2 tests fail
add a name nothing writes                  -> 2 tests fail
make the row answer null unconditionally   -> 19 tests fail
a write via a `patient_info` parameter     -> 1 test fails
`setattr` / an alias                       -> 1 test fails
a `setattr` with a computed name           -> 1 test fails
```

## Conditions corrected, and two bugs found reading them

Four condition strings were wrong or imprecise; `codex review` and the fresh-context review found overlapping sets.

The `later_*` group said *"cleared unless prior therapy is 'More than two lines'"* — inverted for any value outside the four the code enumerates. `prior_therapy` is a bare `TextField` with no choices, so `'Three lines'` over the API would have been told its data was about to be cleared when nothing touches it.

`flipi_score` now names the five options that actually score; `progression` says both criteria must be **known** and unmet, since an unknown leaves it alone; `postal_code` describes the branch that clears it rather than one that cannot run.

Reading them that closely turned up two real defects, filed as **#470**: a zero coordinate is dropped by a truthiness check, so a patient on the prime meridian or the equator gets no `geo_point` and the radius filter silently does nothing; and an explicit `null` for creatinine units reaches an unguarded `.lower()`, which is a 500 on every patient-context endpoint.

## Scope, said out loud

The register covers the **derivation** stage. `ctomop_adapter` derives `prior_therapy` and `metastatic_status` from CTOMOP columns on the `?person_id=` path, and `resolve._coerce_*` drops malformed items from the JSON list fields. Neither is covered — that is **#471**, and `normalize.py` carries a SCOPE note pointing there rather than letting the key imply it speaks for the whole pipeline.

---

Backend: 1023 passed, 6 skipped.

Both halves of the review gate ran. The fresh-context review verified all 36 entries by reading each write site *and* empirically, by subclassing `PatientInfo` with a `__setattr__` tracker over seven input shapes; it reproduced every count independently and returned the `never-stored` gap, the walker hole, the scope question and four smaller corrections. `codex review` found four condition inaccuracies across two passes, two of which became #470.


LeoMo42:
Closing as a duplicate: #449 was implemented and merged into `cb-like-trials` while this branch was in review, in [a868c7b's history](https://github.com/healthkey-ai/exact/commits/cb-like-trials) — `RECOMPUTED_ATTRIBUTES` in `normalize.py`, an AST test holding it to the module, and `upatientRecomputed` / `patientFieldRecomputed` on the two rows.

I branched from `c690136`, which did not contain it, and found out by checking a *different* worktree cut from a newer base. Two independent implementations of the same ticket, same week — worth noting for whoever schedules this backlog.

The two arrived at the same 36 fields and the same AST-comparison design, which is some comfort. Three things differ, and they are now a focused follow-up in **#473** rather than a competing rewrite — it keeps the merged names and adds only what is missing:

1. **`abnormal_kappa_lambda_ratio` and `meets_meas_or_bone_status` answer `upatientRecomputed: false` on the merged code** — measured. Both are `@property` with no column and no setter, so a client reads "this value is the reader's" over a field an edit cannot reach at all. That is the false negative this ticket exists to remove, one stage further along. My own first draft had exactly the same hole; review caught it there, which is the only reason I knew where to look here.

2. **The condition the ticket asked for.** #449 says it plainly — *"Conditional ones want the condition named too"*. 23 of the 36 are conditional, so the boolean reports `mipi_risk` as recomputed for a breast-cancer patient whose `mipi_risk` nothing touches, hiding a control from the one reader who could have used it. Same false positive the ticket names on `meets_gelf`, pointing the other way.

3. **The AST walker matches one name.** It requires the target to be `pi`; the sibling `patient_info_attributes.py` spells the same object `self.patient_info`, so a normaliser copied from there and renamed writes invisibly. Measured: a helper taking `patient_info` leaves the file green, and so do `setattr(pi, …)` and `alias = pi`.

Nothing here is a criticism of the merged work, which is sound and shipped first. Filed from reading it closely: #470 (two defects in the conditions' underlying code — a zero coordinate dropped, a null creatinine unit 500ing) and #471 (the register covers the derivation stage only; `ctomop_adapter` and `resolve` also write patient values).

PR #483, opened 2026-09-14
Preserve PRomop TP53 unknown and positive results through matching

### LINKED ISSUE #481: Preserve explicit PRomop TP53 unknown/positive aggregates through normalization and matching
PRomop 1.3 release gate healthkey-ai/promop#1315. The PRomop approved aggregate emits true or null. EXACT currently recomputes inline/source results using its legacy marker-only predicate and can turn null into false or discard an explicit positive. Preserve an explicitly supplied nullable Boolean through inline snake/camel case input, source-row adaptation, normalization and matching. Keep legacy derivation for omitted fields unchanged. Validate unknown does not satisfy a negative TP53 criterion and positive behavior is retained. No new clinical negative/del17p rule or authorization change.

adamblum:
An explicit PRomop `tp53_disruption: null` was normalized into `false` by EXACT’s legacy marker predicate, and an explicit positive could also be discarded. Preserve the supplied nullable Boolean through inline snake/camel-case input, source-row adaptation, repeated normalization and the attribute service used by eligibility matching. Omitted aggregates retain their existing legacy derivation.

Regression coverage exercises null/true/false, malformed values, source-row adaptation, and required/excluded TP53 criteria. Unknown remains an unresolved candidate criterion rather than a confirmed negative match. Full GitHub CI passed: build and all three backend test groups (run 34853030494). Local execution was unavailable because GDAL is not installed.

Closes #481. Required by healthkey-ai/promop#1315 for the 1.3 release.


LeoMo42:
@adamblum ,

I think you did say that all the changes should go to `dev` branch on EXACT. as we want to keep `main` frozen, no?

PR #488, opened 2026-09-14
Preserve explicit PRomop TP53 unknown results on dev

adamblum:
PRomop's explicit TP53 aggregate can be `true` or `null`. EXACT's `dev` branch currently drops an explicit null during source adaptation and recomputes the value from legacy markers, which can turn unknown evidence into a negative result or lose an explicit positive.

Forward-port #483 to `dev`: preserve an explicitly supplied Boolean/null through inline snake/camel-case inputs, source adaptation, repeated normalization and eligibility matching. Invalid non-Boolean values are rejected. Omitted aggregates keep the existing legacy derivation. Unknown remains an unresolved eligibility criterion and cannot become a confirmed negative match.

Includes 19 regression cases covering normalization, adaptation, validation and required/excluded matching. The original main-branch change passed all three backend CI groups; this PR runs the dev branch's CI independently. The cherry-pick applied cleanly and the diff was reviewed.

Requested handoff to @LeoMo42: review and choose the deployment path. Leave this PR open for that review; no service branch changes or deployment actions are part of this PR. The Render services currently use `dev` and `fix/legacy-trials-schema`, so merging the earlier main PR alone did not update either deployed consumer.

Refs #481, healthkey-ai/promop#1315. PRomop's release gate remains open until compatible consumer deployment is verified.


LeoMo42 (changes requested):
Reviewed as requested. The port itself is clean and the tests are real — but I'm requesting changes, because I think this ships a second silent clinical error in the opposite direction from the one it fixes.

## Blocking: an explicit `null` now overrides positive TP53 evidence in the same payload

A patient with documented `del17p13` + `tp53Mutation` stops being excluded from a trial whose protocol excludes TP53-disrupted patients.

```python
MARKERS = {'cytogenic_markers': 'del17p13', 'molecular_markers': 'tp53Mutation'}

_build_in_memory(dict(MARKERS))                          # dev today
#   tp53_disruption = True   -> vs EXCLUDING trial: not_matched

_build_in_memory(dict(MARKERS, tp53_disruption=None))    # with this PR
#   tp53_disruption = None   -> vs EXCLUDING trial: unknown
```

`unknown` means the trial survives as a candidate. The same holds for the IHC surrogate:

```
p53_ihc=60, key absent       -> True
p53_ihc=60, explicit null    -> None
```

Two things make me think this is the common case rather than an edge:

1. **The PR body says PROMOP's aggregate is `true` or `null`.** If it never emits `false`, then `null` is PROMOP's *not-computed* state, not a clinician's assertion of "unknown" — and the code treats the two as identical.
2. **`tp53_disruption` is a field EXACT derives** (`normalize.py:48`). `ctomop_adapter.SKIP_COLUMNS` already has a category for exactly this — `# Computed by EXACT (do not override): 'bmi'`. This change moves the field the other way, making a source `null` authoritative over our own derivation.

Worth flagging: this was validated on `main`, where the derivation has three evidence sources. `dev` has four — `p53_ihc` (`patient_info_attributes.py:634-636`) doesn't exist on `main`. The blast radius here is strictly larger than the branch it was tested on.

**Suggested shape** — explicit values win, but `null` never discards positive evidence:

```python
provided = getattr(self.patient_info, '_provided_tp53_disruption', _SENTINEL)
if provided is not _SENTINEL and provided is not None:
    return provided              # explicit true/false wins outright
derived = <existing marker / IHC derivation>
if derived is True:
    return True                  # local positive evidence is never discarded
return None if provided is None else derived
```

That keeps every one of the 19 cases except the two `[None]` params where markers are also present — which are the ones currently giving the wrong answer. The marker-positive-plus-explicit-null case has no coverage today, which is why 19 green tests didn't catch it.

**Question that decides the design:** what does `null` on that column actually mean in PROMOP — "clinician recorded unknown" or "we didn't compute it"? Worth pinning down and putting in the docstring, because everything above rests on it.

## Blocking-ish: the new `ValidationError` breaks a contract its own function documents

`resolve.py:154-161`. Over the real HTTP surface:

```
tp53Disruption='true' -> 400      tp53Disruption=1 -> 400      tp53Disruption=0 -> 400
```

Three concerns:

- **It contradicts the function it sits in.** `_build_in_memory` deliberately *coerces* loose input — `_coerce_dates`, `_coerce_numerics` ("CB API can send `\"10.20\"` etc."), `_coerce_json_fields`. The module docstring says of the inline path: *"CancerBot depends on this contract — do not change."* Before this PR those values were harmless, because `normalize` overwrote the field unconditionally.
- **It inverts a stated invariant on the person_id path.** `trials_views.py:76-95` says inline-payload failures are 400 but person_id-path failures must propagate as 500, *"rather than masking it as a misleading 400"*. `ValidationError` is an `APIException`, so `except APIException: raise` re-raises it — a malformed **upstream CTOMOP row** now surfaces as client error.
- **It aborts batch jobs.** `search_trials_for_patients`, `explain_trial_match`, `probe_eligibility`, `compare_trials` all reach this via `build_patient_info_from_ctomop_row`; a DRF `ValidationError` there is a traceback, not a 400.

Either coerce in the spirit of the neighbouring `_coerce_*` helpers, or scope the raise to the inline path and namespace it under `patient_info` like the sibling errors.

## Worth doing now while the seam is open

The same bug exists on five more fields — explicit `null` in, confirmed negative out:

```
renal_adequacy_status          None -> False
hepatic_adequacy_status        None -> False
haematological_adequacy_status None -> False
tnbc_status                    None -> False
metastatic_status              None -> False
```

`pi._provided_fields = set(filtered)` captured once at `resolve.py:153`, plus a shared helper, costs about the same and closes the category instead of one instance. Otherwise each of these arrives as its own PR later.

Also: `_build_in_memory` being the *only* sanctioned construction site is what makes the marker correct, and nothing enforces it. A line in `PatientInfo`'s docstring would be cheap insurance.

## What's good here, and I want to be clear about it

- Single choke point: `grep "PatientInfo("` across `trials/` returns exactly one construction site, and `build_patient_info_from_ctomop_row` ends at `return _build_in_memory(cleaned)`, so the PROMOP path is genuinely covered. (I initially thought it wasn't — I hadn't read to the end of the function. It is.)
- The private attribute survives everything: repeated `normalize`, fresh `PatientInfoAttributes`, the cached `attributes_service`, `copy`/`deepcopy`. It does not leak into serialization (`to_representation` filters `_`-prefixed keys).
- Three-valued semantics are correct: `None` is genuinely unknown to the matcher, via `is_attr_blank()` → `continue` in the SQL prefilter and `'unknown'` in `_match_computed_subattr`. Your claim that unknown can't become a confirmed negative holds.
- **The 19 cases are real.** Reverting each source change independently: `resolve.py` block → 15 failures, the `patient_info_attributes.py` early return → 10, the `ctomop_adapter.py` line → 1. Nothing is dead.
- Cherry-pick fidelity is exact: the test file is byte-identical to `main`, and the other diffs vs `main` are pre-existing `main`↔`dev` drift, not port damage.

## On the deployment question

Nothing here argues for a particular service branch. But I wouldn't push this to any consumer serving real patients until the first item is settled — the Render services running `dev` would inherit that behaviour on the next redeploy.

## Unrelated, pre-existing, worth its own issue

`normalize_ctomop_row` resolves markers by *display title*, so a row carrying codes (`'tp53Mutation'`, `'del17p13'`) resolves to `None` and the legacy derivation silently yields `False` on the person_id / management-command path. `'TP53 Mutation'` works. Not this PR, but it's the same class of silent wrong answer.

LeoMo42 (approved):

PR #512, opened 2026-09-15
Expose the TrialMatches bridge on main, so staging serves it

vtrv101:
`main` exposes `./TrialMatches` and `./types`, but not the bridge, so HealthTree ONE asking for `exact_remote/TrialMatchesBridge` gets a module that is not there and the Clinical Trials screen fails to mount.

#414 added it, but it merged into `dev` (it was retargeted from `main` after I opened it). The staging service builds from `main`, so the fix never reached the deployment.

## What this is

The three federation commits from #414, cherry-picked onto `main` with `-x`, and nothing else. The four commits that PR also carried — from the in-flight `feat/phr-federation` / `fix/legacy-trials-schema` stack — are deliberately left out.

| commit | |
|---|---|
| `109c489` | expose TrialMatches through a framework-agnostic bridge |
| `907c4fa` | resolve the signed-in patient inside the bridge |
| `c9ee45a` | keep injected CSS out of the host's utilities layer |

Seven files: the bridge component, `cssLayer` and its test, `injectStyles`, the exposes entry, and the `@module-federation/bridge-react` dependency.

## Verified on this base

`npm ci` and `tsc -b` clean, the four federation suites pass (34 tests), and `npm run build` emits `TrialMatchesBridge-DIggy6if.js` into `dist/remote/assets` — so the module the host asks for exists in the artifact.

## Note

`dev` and `main` have diverged here and this leaves the same change on both. If the intent is for `dev` to merge into `main` wholesale, this becomes a duplicate — say so and I will close it in favour of that.

https://claude.ai/code/session_015dVAjtcAhgRnzmzFpYgWww


LeoMo42 (commented):
Verified the premise and the artifact locally, because **CI does not cover this PR at all** — see below. The change does what it says; three notes, none of them a reason to hold staging broken.

## What I checked

| claim | result |
|---|---|
| `main` lacks the bridge, `dev` has it | ✅ `main` exposes only `./TrialMatches` and `./types`; `TrialMatchesBridge.tsx` absent |
| cherry-picks are genuine | ✅ all three carry `-x` trailers, and the originals `9e02a8e` / `e85e4ad` / `32ece56` are ancestors of `origin/dev` |
| the picked bridge is self-contained on `main` | ✅ 137 lines, imports only `react`, `axios`, `bridge-react/v19`, `./TrialMatches`, `./types` — all present on this branch |
| `tsc -b` clean | ✅ exit 0 |
| tests pass | ✅ 4 files, 34 tests |
| the host's module exists in the artifact | ✅ `npm run build` emits `TrialMatchesBridge-BwNRIAU8.js`, and `virtualExposes-Cl5Ik1I9.js` declares `./TrialMatchesBridge` and imports that chunk |

The last row is the one that matters, so I checked it at the artifact level rather than trusting the file list.

## 1. The green checks are not evidence for this PR

All seven files are under `frontend/`, and `.github/workflows/django.yml` has no `npm` or `frontend` step — the pipeline builds and tests the Django side only. So `build` + `test (1..3)` passing says nothing about this change. That is why I ran the frontend locally.

Not a request to change anything here; it does mean a frontend regression on `main` cannot be caught by CI today, which seems worth its own issue.

## 2. The branch has four commits, not three

The description says "the three federation commits from #414, cherry-picked onto `main`, and nothing else". There is a fourth, `2d21b04` — the loading copy, "Loading your health profile…" → "Finding your trial matches…". It is a one-line change with a clear rationale in its own message, and I have no objection to it; the description just no longer matches the branch.

It also explains a detail that would otherwise look wrong: the built chunk here is `TrialMatchesBridge-BwNRIAU8.js`, not the `TrialMatchesBridge-DIggy6if.js` the description quotes. Different content, different hash — consistent with the fourth commit landing after that paragraph was written.

## 3. The bridge arrives on `main` without its own tests

Three test files exist on `dev` and not here:

```
frontend/src/federation/TrialMatchesBridge.test.ts
frontend/src/federation/bridgeState.test.ts
frontend/src/federation/nctLink.test.tsx
```

So the 34 passing tests do not include a single test of the component this PR exists to ship. That is a consequence of picking three commits rather than the later work, not a mistake in the picking — but it should be said out loud, since "34 tests pass" reads as coverage of the bridge and is not.

Related: `TrialMatchesBridge.tsx` differs from `dev`'s by 424 lines and `dev` has gained `bridgeState.ts`. A later `dev` → `main` merge will conflict in that file, which is the duplicate risk your own Note raises. My read: land this now, because staging serving a missing module is the worse state, and resolve the divergence when the wholesale merge happens — taking `dev`'s version, which is the one with the tests.

Not approving formally since the merge call is the maintainer's, but I found nothing blocking.

LeoMo42 (commented):
**Correction to my review above — point 1 was wrong.**

I wrote that "the green checks are not evidence for this PR" because
`django.yml` has no frontend step. It does not, but `.github/workflows/frontend.yml`
does, and it ran here: the check list is `backend / build`, **`frontend / build`**,
`backend / test (1..3)`. That job runs `npm ci`, `npm run typecheck`, `npm test`
and `npm run build` on every PR touching `frontend/**`. I saw two rows named
`build`, assumed they were both backend, and grepped one workflow file instead
of the directory.

So CI does cover this PR, and my local run was a duplicate of it rather than a
substitute. The local results stand and still agree: `tsc -b` exit 0, 4 files /
34 tests, and the artifact declares `./TrialMatchesBridge` in
`virtualExposes-Cl5Ik1I9.js`.

Points 2 and 3 are unaffected: the branch has a fourth commit the description
does not mention (`2d21b04`, the loading copy, which is also why the built hash
differs from the one quoted), and the bridge's own tests
(`TrialMatchesBridge.test.ts`, `bridgeState.test.ts`) stay on `dev` rather than
arriving with it.

What survives from point 1 is narrower, and I have moved it to #514: CI proves
the remote builds, not that it still exposes the modules hosts import — dropping
an `exposes` key builds green, which is exactly the shape of the breakage this
PR repairs.

Sorry for the noise.

